PR #520, opened 2022-08-15
improve(home): interactive scheme

unordinarity:
## Changelog

replaced static scheme in home page with interactive one

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -4,8 +4,14 @@ module.exports = {
         browser: true,
         es6: true,
     },
-    extends: ["@eslint-kit/patch", "@eslint-kit/base", "@eslint-kit/react", "@eslint-kit/prettier"],
-    parser: "babel-eslint",
+    extends: [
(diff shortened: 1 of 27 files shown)


illright:
image
Something doesn't seem right here

unordinarity:
@illright 
> Something doesn't seem right here
What a pain, I forgot we have two color themes

azinit:
@illright if u have another suggestions about scheme, u're welcome)

illright:
I think that we shouldn't have `processes` in that scheme. It's pretty vague and since we don't have any examples in our gallery that use processes, I imagine it would confuse people

azinit:
> @illright if u have another suggestions about scheme, u're welcome)

Also, if u have ideas about scheme styles and general UI, u can send figma/screens for help

azinit:
@unordinarity Ну и поправить бы ошибку линтера и конфликты зарезолвить

illright:
Closing as stale

PR #525, opened 2022-09-06
WIP: Setup IndexPage for docs

azinit:
## Changelog

Setup IndexPage docs with simplifying "OVerview" article

image

image

image

https://www.gatsbyjs.com/docs/


illright:
Okay, here goes.

### 1. The actual textual content is too far down the website.

Here's my flow to start learning about FSD with this docs setup (excuse the slightly sarcastic language, I'm imitating a busy user):
1. Click **Get Started** on the landing page
2. Read the introduction paragraph
3. Look through the cards on the index page (they're pretty big, on my 14" laptop I need to scroll a lot to get an overview)
4. Click **Get Started** (again? I though I was already getting started)
5. Read another introduction paragraph (again? give me the content already)
6. Look through the cards:
  - "overview", finally; 
  - "quick start", after two clicks on "get started"? rrrrrrrrrrrr;
  - "decomposition cheatsheet", will probably refer to it later, don't need it for now
7. Click "overview"
8. Aha, readable text!

Yeah, I think it's a bit too hidden away for being so essential. Here's the Gatsby flow, and why they're successful in their structure, in my opinion:
1. Spend literal ages finding the link to the documentation from their godforsaken landing page (for heaven's sake Gatsby do something about your fricking landing page)
2. Read the introduction paragraph
3. Click on a colorful and inviting card that's prompting to start
4. Aha, readable text!

So, what they got right is:
* their card is super cute and inviting
* the cards are horizontal ⇒ you see all the options right in front of you, no need to scroll even on a laptop
* the card that gets you started actually gets you started instead of leading to another index page

### 2. The edits to the content of the intro make it harder to comprehend (to me personally)

> Also, FSD have few core-concepts:

This section, I remember it from the original intro page (and remember removing it). What I don't like about it is the imparallelism of the options listed there:
* Public API — that's a visible attribute of an FSD project, represented by actual files (`index.ts`)
* Isolation — that's an invisible concept that is more of a defining principle for why things are the way they are in the methodology
* Domain Driven — that's a property of FSD, syntactically an adjective (as opposed to isolation, which is syntactically a subject)

So hopefully you can see where I'm going with this — this list feels a bit haphazard in its content, which confuses me as a reader as to what exactly it's trying to tell me. And links, the links are back, nooooooo :D

> `└── src/`

This huge file system tree thing is nice, but not very suitable for the introduction, imo. The introduction is meant to hammer the following things into the mind of a reader:
* FSD is a frontend methodology and it's useful to me because I'm a frontend developer in a team
* FSD is telling us how and where to put files (the particulars of this don't matter quite yet)
* The three knights of FSD are layers, slices and segments
* I can tell these things to my teamlead to make them adopt the methodology
* I can take these steps to adopt FSD incrementally

This filesystem feels like too much details, so I think we'd better wait with it. Also I think it'd be really nice as an interactive widget, not just a code block, but that's 💄 for the future.

> See also

This is kind of a crossroads for the reader. After reading the intro (till the end), they are slightly interested in the methodology, so now they want to learn the particulars. I think it's best at this stage to lead them by the hand to the next page exactly instead of confronting them with a choice.

### 3. When the very first thing in the docs starts with a caution, I'm immediately worried

We have this prominent caution block at the very start of the overview page. Me as a reader, I'm used to seeing such things in alpha-stage products, and that caution is usually telling me to expect bugs and be careful in production. We, however, don't really need our users to watch out for anything, and in my opinion it's expected that you won't understand everything about the methodology by simply reading one page out of many that there are. So I'm not sure what this warning is trying to achieve, but it instills anxiety in me as a reader.

### 4. [minor] Not a fan of those bright blue rectangles around the terms

They are too bright, they hurt my eyes a bit. And also I'm not quite sure why we need to highlight these terms **so** prominently, they are important, but they're not that important.

---

Now here's my list of suggestions:

1. Style the cards in the index page differently. Make them fun and colorful, like in Gatsby, arrange them in a grid and cut down the amount. 2 cards is the ideal amount, honestly :)
2. Roll back the edits to the intro. Yeah, I'm definitely saying this because I wrote the previous version, but I truly believe we had it quite nicely already, I don't really see a reason to change it

src/features/header/ui.jsx
```diff
@@ -14,7 +14,7 @@ export function Header() {
                 <h1 className="hero__title">{siteConfig.title}</h1>
                 <p className="hero__subtitle">{translate({ id: "features.hero.tagline" })}</p>
                 <div className="button-group">
-                    <Link className="button button--primary button--lg" to="/docs/intro">
+                    <Link className="button button--primary button--lg" to="/docs">
```
azinit, line 17: > 1. The actual textual content is too far down the website.

Mostly agree, but what if we do next:
- GET_STARTED button => `/docs/get-started/overview`
- Docs navlink => `/docs`

On my opinion we solve your remarks:
- we omit redundant IndexPages navigating and navcards clicking
- we go to "readable text" at one click 🎊

azinit: Navcards also could be simplified (we have #445  issue)

i18n/en/docusaurus-plugin-content-docs/current/get-started/overview.md
```diff
@@ -0,0 +1,172 @@
...
+
+Each slice, in turn, consists of <mark>segments</mark>. These are tiny modules that are meant to help with separating code within a slice by its technical purpose. The most common segments are `ui`, `model` (store,  actions), `api` and `lib` (utils/hooks), but you can omit some or add more, as you see fit.
+
+:::note
+
+In most cases, [it is recommended][ext-disc-api] to place `api` and `config` only in the shared layer
+
+:::
+
+**Also, FSD have few core-concepts:**
```
azinit, line 59: > 2. The edits to the content of the intro make it harder to comprehend (to me personally)

> Also, FSD have few core-concepts

Partially understand, but you're welcome to suggest your version of this part about concepts :)

azinit: > This filesystem feels like too much details, so I think we'd better wait with it. Also I think it'd be really nice as an interactive widget, not just a code block, but that's 💄 for the future.

Again, suggest your vision, how we could simplify files-structure for Overview :) 
(or where we should place this structure to easy-to-find user-cases)

azinit: > See also

Agree, we could omit this part

i18n/en/docusaurus-plugin-content-docs/current/get-started/overview.md
```diff
@@ -0,0 +1,172 @@
+---
+sidebar_position: 1
+---
+
+# Overview
+
+:::caution
```
azinit, line 7: > 3. When the very first thing in the docs starts with a caution, I'm immediately worried

Again, its legacy from old article

We could omit it, agree

i18n/en/docusaurus-plugin-content-docs/current/get-started/overview.md
```diff
@@ -0,0 +1,172 @@
...
+- A very simple app of a single page might not need the benefits of FSD and suffer from the overhead. However, FSD promotes a nice way of thinking, so feel free to use on the tiniest projects if you want.
+- A huge app, the size of the Google Cloud admin dashboard, will require a custom architecture. It could still be based on FSD, by the way.
+
+FSD doesn't enforce a particular programming language, UI framework or state manager — bring your own or see some [examples][refs-examples].
+
+If you have an existing project, fear not — FSD can be adopted incrementally. Just make sure that your team is **in&nbsp;pain** from the current architecture, otherwise a switch might not be worth it. <!-- For migration guidance, see the Migration section. TODO: add the migration section and link to it here -->
+
+## Basics
+
+In FSD, a [project consists][refs-splitting] of <mark>layers</mark>, each layer is made up of <mark>slices</mark> and each slice is made up of <mark>segments</mark>.
```
azinit, line 29: > 4. [minor] Not a fan of those bright blue rectangles around the terms

image

I tried different variants, but highlighting by mark at this place - looked good for me
(with comparing of default styles, _italic_, `code` or **bold**)

Also as you said - it need for accent basic abstractions of FSD
And basic content visual splitting for easy reading
(i'm not sure that it so "hurts" 🤷‍♂️ )

@feature-sliced/core What do u think?
(especially - @unordinarity @Rin-Akaia-eth)

i18n/en/docusaurus-plugin-content-docs/current/intro.mdx
```diff
@@ -98,7 +19,7 @@ import Link from "@docusaurus/Link";
     description="A tour over the basic concepts and structure as well as a comprehensive review of a React sample project"
     to="/docs/get-started"
     Icon={"🚀"}
-    tags={['Basics', 'Quick start', 'FAQ']}
+    tags={['Overview', 'Quick start', 'FAQ']}
```
azinit, line 22: > Style the cards in the index page differently. Make them fun and colorful, like in Gatsby, arrange them in a grid and cut down the amount. 2 cards is the ideal amount, honestly :)

Again, we have #445 issue for this

Redesigning navcards and docs sections - it more difficult task 🤔 
(unless you have "easy to fix/impl" suggestions)

azinit: Also, we couldn't decrease amount right now because:
- We have incomplete restructuring of doc (#263)
   (we will discuss it detailed at next DOCS sync)
- Without dropdown, this page will be the only one entry-point for docs navigating
   (it is why its `IndexPage`)

i18n/en/docusaurus-plugin-content-docs/current/intro.mdx
```diff
@@ -1,94 +1,15 @@
 ---
 sidebar_position: 1
+slug: /
 pagination_next: get-started/index
 ---
 
-# 🔎 Intro
```
azinit, line 6: > Roll back the edits to the intro. Yeah, I'm definitely saying this because I wrote the previous version, but I truly believe we had it quite nicely already, I don't really see a reason to change it

Its more complicated question, yeah 😄 

image

According our last discussion about docs problems (screenshot above), we decided to setup "Table of Content" page as IndexPage of whole documentation. Also we solved that for this we should disassemble Intro: smth to landing, smth to GetStarted

**And my main point** - we have TOO MANY sections for docs, and Intro (that should help to "get start") have similar responsibilities as `/get-started` section

As a result, we have:
- less redundant sections
- refactored and single "Overview" article on `/get-started` (next, we'll simplify access it like "one-click")
- simple "IndexPage" of docs (a.k.a ToC) with possibilities to refine
   (now it not depend from old "Intro" content)

azinit: @unordinarity @illright @Imperyall @Rin-Akaia-eth What do u think?

i18n/en/docusaurus-plugin-content-docs/current/get-started/overview.md
azinit: > (i'm not sure that it so "hurts" 🤷‍♂️ )

UPD: I also tried to overdo with it, of course

Its why I dont highlight another parts with abstractions and concepts 😏

azinit: UPD:UPD: But as compromise - i could omit highlighting for "detailed description"

I mean

```md
<!-- with highlighting -->
In FSD, a [project consists][refs-splitting] of <mark>layers</mark>, each layer is made up of <mark>slices</mark> and each slice is made up of <mark>segments</mark>.
```

```md
<!-- without highlighting -->
The **layers** are standardized across all projects and vertically arranged. ...//

Then there are **slices**, which partition the code by business domain. ..//

Each slice, in turn, consists of **segments** ...//
```

src/features/header/ui.jsx
illright: My concern with this proposal is that this will create a page in the flow that feels "preceding" and that would make me afraid as a reader of missing out.

What I mean is:
If the flow is `Page 1 => Page 2 => Page 3 => ...`, but the link in the landing page sends me to page 3, I will be okay, I will start reading and everything will be fine. That is, until I eventually accidentally stumble onto pages 1 and 2 and start thinking "oh shi-, I think I didn't start at the beginning".

If some part of the methodology doesn't make sense to me, I will be more likely to blame it on myself with the thoughts along the lines of "well I should've read from the beginning, there are probably some ground things and base concepts that I didn't grasp". That way, we might get less users complaining about the docs even in cases when the docs are the problem. 

My personal ideal is not to go to readable text at one click, and it's never really about the clicks, as Steve Krug says in an UX design classic "Don't Make Me Think". What matters is that in our case, with every click comes a decision:
* [on /docs] I see a lot of nav cards, I'll click the first but I need to remember to check out the rest
* [on /docs/get-started] Oh hell, there are more choices here... Clicking the first one, gotta remember...

Again, what makes Gatsby so nice is that they have very few options and the every option gets you started right away. So I suggest to keep the landing page link to lead to `/docs`, but change the links of the nav cards in the `/docs`

i18n/en/docusaurus-plugin-content-docs/current/get-started/overview.md
illright: > Partially understand, but you're welcome to suggest your version of this part about concepts :)

But why do we need it?

i18n/en/docusaurus-plugin-content-docs/current/get-started/overview.md
illright: > it need for accent basic abstractions of FSD

But is there a real problem being solved here? I don't think there are any complaints that the users are struggling to understand that FSD is really about these three things, especially when they are doubled with a visual diagram which represents the things and nothing more. I think we don't need more highlighting here, italics should work fine

i18n/en/docusaurus-plugin-content-docs/current/intro.mdx
illright: > Without dropdown, this page will be the only one entry-point for docs navigating

Notice that the Gatsby index page is also quite packed with links and items, but it works, because only two of them are heavily prominent, the rest is just quiet links below. I suggest we do the same, clearly delineate the priorities

illright: > Redesigning navcards and docs sections - it more difficult task 🤔

Agreed, it's not as simple. However, I think that if we don't do it, but still apply this PR, we will be in a risky transitive state where the navigation may actually be worse for a while. So maybe we could do them together as a unit of work, with a bit more planning and time devoted to it?

i18n/en/docusaurus-plugin-content-docs/current/intro.mdx
illright: It's hard for me to say here what do I think) I agree that we have too many sections and that some of them have similar responsibilities.

I just don't really understand what the comment in this thread refers to in this PR. What change is it describing?

i18n/en/docusaurus-plugin-content-docs/current/get-started/overview.md
azinit: > But why do we need it?

Because it's not obvious for users at start

And **little note** about concepts would be useful for high-level understanding 🤷‍♂️ 

(only "layers-slices-segments" concept is not enough, even for "Overview")

src/features/header/ui.jsx
azinit: Lets discuss on next call

i18n/en/docusaurus-plugin-content-docs/current/get-started/overview.md
azinit: > But is there a real problem being solved here? I don't think there are any complaints that the users are struggling to understand that FSD is really about these three things, especially when they are doubled with a visual diagram which represents the things and nothing more. 

For example

image

azinit: >  I think we don't need more highlighting here, italics should work fine

@Rin-Akaia-eth @unordinarity @Imperyall

i18n/en/docusaurus-plugin-content-docs/current/intro.mdx
azinit: > I suggest we do the same, clearly delineate the priorities

If u suggest these navcards - I don't mind

image

But I think user also could be "confused" as u say, if we suggest him more choices and we'll again get effect "make me afraid as a reader of missing out." 😏

i18n/en/docusaurus-plugin-content-docs/current/intro.mdx
azinit: > I just don't really understand what the comment in this thread refers to in this PR. What change is it describing?

Ok, lets discuss at next call 🤷

i18n/en/docusaurus-plugin-content-docs/current/get-started/overview.md
illright: But why do users need to know this from the start? I can only speak for myself, but I don't think I'd want to see this info in the overview.

Public API — this is something that I would learn in the next page, "applying FSD in practice", when I would create an index file. The docs would say "this file is called a public API of the module, and it's an important part of isolation".

Isolation — kinda the same as public API. You don't need to know a lot of background knowledge to create an index file, so this is something that can be deferred, in my opinion

Domain Driven — yeah, it's important, but I think I'd rather it be mentioned when we introduce users to slices on that page, so that it would be in context

i18n/en/docusaurus-plugin-content-docs/current/intro.mdx
illright: I should point out that this effect of fear of missing out only happens when the options appear equally important, for example, when they are visually the same. When there are two prominent options and the rest is very quiet, I'm only concerned about the two prominent ones and the rest I label in mind as "reference material when I'll have exact questions"

i18n/en/docusaurus-plugin-content-docs/current/get-started/overview.md
Rin-Akaia-eth: I like these rectangles, but I definitely would like to use another color for them, something much more neutral

azinit: Then we could use `code` snippets - they have same styles but with grayscale

But it intersect with layers examples
(`shared` `entities`... and `layers` `slices` `segmebnts`)

Also can suggest to save "blue rectangle" but with less saturation (more gray, but not absolute grayscale)

@RinAlaiaEth @illright what do u think?

illright: I'm personally against any background for these. I could accept colored text (but we need to be careful not to make them look like links), but the background adds too much excess emphasis, imo

Rin-Akaia-eth: @azinit, @illright, I just realized that I would actually make every methodology concept mention to be a link to the page about it ( wikification? )

Rin-Akaia-eth: It will also resolve a highlighting problem

illright: I'd object to having them as links as well. It is indeed wikification, and what works for the wiki doesn't necessarily work for the introduction page. Since the start of my work on the docs I've been on a quest to decrease the amount of links in the introduction to an absolute minimum

Rin-Akaia-eth: @illright why do you think it's a problem?

azinit: fixed

i18n/en/docusaurus-plugin-content-docs/current/intro.mdx
azinit: fixed

i18n/en/docusaurus-plugin-content-docs/current/intro.mdx
azinit: fixed

i18n/en/docusaurus-plugin-content-docs/current/get-started/overview.md
azinit: fixed

i18n/en/docusaurus-plugin-content-docs/current/get-started/overview.md
azinit: fixed

src/features/header/ui.jsx
azinit: fixed

PR #536, opened 2022-11-24
Add used import at documentation/widgets.md

lopezac:
Add Navigation feature used in line 48, but not imported.


i18n/en/docusaurus-plugin-content-docs/current/reference/units/layers/widgets.md
```diff
@@ -37,6 +37,7 @@ And if the first option is easy to implement, then everything becomes not so tri
 
 ```tsx title=widgets/header/ui.tsx
 import { SearchBar } from "features/search-bar";
+import { Navigation } from "features/navigation";
```
azinit, line 40: Thx for your care! Could u please add same import on ru version of article?

lopezac: Yes I'll do it.

azinit (approved):
Thx for contribution!
We'll glad to see u again 🚀

PR #538, opened 2022-11-24
Add english link documentation/index.md

lopezac:
Add english link to wikipedia article about GRASP (object-oriented design), because there is only russian version linked.


i18n/en/docusaurus-plugin-content-docs/current/reference/isolation/index.md
```diff
@@ -65,7 +65,8 @@ Violation of this rule is a signal of problems:
 - [(Guide) About achieving low coupling][refs-low-coupling]
 - [(Discussion) Coupled entities](https://github.com/feature-sliced/documentation/discussions/49)
 - [(Discussion) About cross-imports and analysis зависимостей](https://github.com/feature-sliced/documentation/discussions/65#discussioncomment-480822)
-- [**GRASP** Patterns](https://ru.wikipedia.org/wiki/GRASP)
+- [**GRASP** Patterns (russian)](https://ru.wikipedia.org/wiki/GRASP)
```
illright, line 68: question: if this is the English version of the page, do we really need a Russian link here?

suggestion: keep only the Russian link in the Russian translation and only the English in the English one

lopezac: Yes you are right.

illright (approved):

unordinarity (approved):

PR #539, opened 2022-11-24
Fix name collisions documentation/tutorial.md

lopezac:
Names collisions should be resolved at the level of public interface, not implementation. Here it exports directly FooCard, but it should export Card as FooCard, the same with FooThumbnail.
Source: https://feature-sliced.design/docs/reference/public-api#resolution-of-collisions


illright (approved):

unordinarity:
@lopezac that's a good point, thank you for noticing

unordinarity (approved):

PR #541, opened 2022-11-26
Update auth guide path typos

lopezac:
In the code examples it talks about importing viewer model, but in the files path it says "entities/user/...", so I think and comprehend it talks about viewer, so I changed that.
Change 'redux' to 'Redux' when necessary.


lopezac:
I added commits from another pull request, I tried to delete the commit, but I force-pushed and cant open it again

PR #543, opened 2022-12-09
Update segment ui examples, in both en/ru

lopezac:
* Add used import "Layout"
* Add final ;
* Add in the tree, the styles.module.scss file used in the examples below.

I think adding the styles.module.scss file in the tree helps understanding that inner modules use the styles defined at the start module.
Also the Layout import may help someone that does not understand where it comes from.

### DIFF
--- a/i18n/en/docusaurus-plugin-content-docs/current/reference/units/segments.md
+++ b/i18n/en/docusaurus-plugin-content-docs/current/reference/units/segments.md
@@ -75,10 +75,12 @@ It can contain inside:
     |   ├── content/
     |   |     ├── sort/
     |   |     └── table/
-    |   └── index.tsx/
+    |   ├── index.tsx
+    |   └── styles.module.scss
 ```
 
 ```tsx title={layer}/{slice}/ui/index.tsx
+import { Layout } from "shared/ui";
 import Toolbar from "./toolbar";
 import Content from "./content";
 import styles from "./styles.module.scss";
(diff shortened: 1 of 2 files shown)


i18n/en/docusaurus-plugin-content-docs/current/reference/units/segments.md
```diff
@@ -75,10 +75,12 @@ It can contain inside:
     |   ├── content/
     |   |     ├── sort/
     |   |     └── table/
-    |   └── index.tsx/
+    |   ├── index.tsx
+    |   └── styles.module.scss
 ```
 
 ```tsx title={layer}/{slice}/ui/index.tsx
+import { Layout } from "shared/ui";
```
illright, line 83: issue: I don't think it was intended for `Layout` to come from `shared`, I think it's a segment-local component, `./Layout.tsx`

illright (approved):
I'm a bit concerned about the `Layout` thing, but honestly this page is so old and lacking so much useful information that I really don't think it's even worth improving. We can just merge this as is.

Thanks for looking out for these inconsistencies!

i18n/en/docusaurus-plugin-content-docs/current/reference/units/segments.md
lopezac: So you say that it should also appear in the [tree "Complex UI for the layer"](https://feature-sliced.design/docs/reference/units/segments#complex-ui-for-the-layer) alongside index and styles.module?

illright: Yes, I think that's how it was intended, but my guess is as good as yours. We may wanna find the original author of this page to see what they meant.

Rin-Akaia-eth (approved):

azinit:
Thanks for contribution! :heart:

PR #551, opened 2023-01-24
Refactor/knowledge types

xxtereshko:
## Changelog

See diff.


i18n/en/docusaurus-plugin-content-docs/current/about/understanding/knowledge-types.md
```diff
@@ -3,28 +3,30 @@ sidebar_position: 3
...
-# Types of knowledge in the project
+# Knowledge types in the project
 
-In any project, you can distinguish certain types of knowledge
+The following "types of knowledge" can be distinguished in any project:
 
-- **Fundamental knowledge**  
-  Knowledge that practically does not change over time.  
-  *Algorithms, computer science, the mechanisms of the YAP and its API, etc.*
+### Fundamental knowledge
```
illright, line 10: question: why did you decide to use headings instead of a list?

In my opinion, this page should not exist in the form that it currently exists in, and instead it should be a part of a larger page discussing the benefits of FSD. With that in mind, I'd actually prefer to shorten this page as much as possible so that it wouldn't occupy too much space when it's merged into another page.

xxtereshko: I think we can refactor itI think we can rewrite it when we paste it into another page. In general, lists with large items are harder to read.

illright: I agree about the large lists. However, I think we can approach this from another direction by reducing the text content (I'd argue, without losing too much informational value).

I took the liberty of experimenting and came up with this, what do you think about it?

image

I changed it back into a list and reworded the description of the third point to be shorter

xxtereshko: Looks good

illright: Introduced the change, let me know if there's anything else you'd like to tweak, otherwise we can merge this

illright (approved):

PR #552, opened 2023-01-26
Fix anchor in 1.1 header

barabaiiika:
## Background

This PR fixes the missing anchor in the 1.1 header:  
  
Снимок экрана 2023-01-26 в 07 42 47


azinit (approved):
Thanks for contribution! ❤️

PR #560, opened 2023-03-26
Added react-native example to documentation

berdimyradov:
I created a ReactNative (Expo) example of Todo-App based on FSD 

Please consider this PR in order to add one more example to documentation


illright:
Hi! Thanks for the contribution! This one will be particularly valuable since we didn't have React Native projects previously.

I have a few remarks about the use of FSD in your project:
* issue: I noticed that your `pages/` directory contains an index file that exports everything from 3 pages. FSD prescribes to have index files on the _slice_ level (and, optionally, on the segment level). Also, it's not clear from the `pages` directory what the slices are. The same applies to most of the other layers, I think it's important to clearly delineate slices and segments in projects.
  suggestion: use the following directory structure:
  image
  In this case, the three slices are `add-task`, `list-tasks`, `view-task-details`. Each slice has an index file and a `ui` segment in form of a single file. Note that this creates an unfortunate stack of `ui.tsx` files in your code editor's tabs, so to overcome that, you can name the UI files `ui/add-task-screen.tsx` instead. More verbose, but clearer.
* issue (non-blocking): you have a `hooks` segment in `entities/task`. The name of this segment is a bit problematic because it's not obvious from the name what the hooks inside are for. For example, a hook like `useModalVisibleState` is for UI (therefore, should be in the `ui` segment), while a hook like `useCompleteTask` that exports a mutation is for `lib`, since it represents the logic of the feature. In your case, the hooks `useTaskDispatch` and `useTaskSelector` should probably belong to `lib`. However, since in Redux, these hooks apply to the whole store, not just a particular reducer, they should probably be kept in `shared`
* issue: some of your features don't contain interactinoal business logic, for example, `task-footer`. That is a presentational component for the `task` entity, it should be placed in `entities/task/ui`. The same goes for `task-summary`, that doesn't seem like a feature to me, the button for adding a new task should be placed in `features/task-add/ui`, and the rest can probably go to `pages` or `widgets` with a slot for the said button to add a task.
* praise: you have no forbidden imports, great job at separating the code into layers and features!

Feel free to reach out if you need any help or guidance about those

berdimyradov:
@illright Hi. Thanks for your review. Great feedback 👍

Please take a look to the updates according to your issues:

- Commits related to issue 1
	- [Refactored `/pages` into slices](https://github.com/berdimyradov/fs-rn-todo-app/tree/6b697f203b041d7e23c21dfa15466b6b607ecdca)
- Commits related to issue 2
	- [Moved `redux-hooks` to `shared` layer](https://github.com/berdimyradov/fs-rn-todo-app/tree/ba186ad2f02f425853b8fe4421e20adca009ef70)
- Commits related to issue 3
	- [Extracted `task` entity into explicit slice](https://github.com/berdimyradov/fs-rn-todo-app/tree/6c21d546c1c13a7246ca5a8c74391c71a0d3084b)
	- [Moved `task-footer` from `features` slice to `entities` slice](https://github.com/berdimyradov/fs-rn-todo-app/tree/61a8f214964c0169b33e29d4e344b333eaa3fddf)
	- [Moved `task-summary` to `entries`](https://github.com/berdimyradov/fs-rn-todo-app/tree/e51735d51bc8985f59932947a1bd8091f4c61af7)
	- [Extracted *task* `widgets` into explicit slice](https://github.com/berdimyradov/fs-rn-todo-app/tree/85e0e6ba0f31bc825048a2a7bf2cfbf036240f34)

azinit:
@illright

illright (approved):

PR #569, opened 2023-04-23
Update the README

illright:
## Changelog

1. Updated the banner to include links and be cool (and support dark mode)
2. Purged the copypasta from the docs, it doesn't need to repeat the website


azinit (approved):
Looks great! Especually, theme-adaptive assets and removing duplicate content

PR #570, opened 2023-04-24
Add React/ReduxToolkit example

noveogroup-amorgunov:
## Background

Add actual react/redux example. I see current react/redux examples a bit outdated and simplified (only with one entity), so I think this example can everyone who want to see a more complex example

- repo: https://github.com/noveogroup-amorgunov/nukeapp
- demo: https://nukeapp.netlify.app/

I will be glad to get project review


illright (changes requested):
praise: really cool app! I love the visualization in debug mode

issue: I noticed that your API segments in entities import from `'@/shared/api/baseApi'`, the import should be from `'@/shared/api'` since `baseApi` is considered internal implementation of the API segment of shared
https://github.com/noveogroup-amorgunov/nukeapp/blob/627700eb47e39064c478633ed157fcadfbbc9e4a/src/entities/product/api/productApi.ts#L1

praise: it's nice that you implemented cross-imports by the RFC, it gives it a practical manifestation
suggestion (non-blocking): add a note to the README that links to that RFC since it's not part of the FSD spec (yet). Just so people exploring the project would have a chance to learn about this non-standard extension to FSD

I briefly looked over the code and everything seemed pretty solid to me, especially the api segments of entities. If you could update the imports from `shared` to use index files on segment level, I'd have no other remarks.

Thanks for the contribution!

noveogroup-amorgunov:
@illright thank you for review. I add public api (index.ts) files for all segments in shared layer (api, ui, model and lib) and add comments about RFC inside `@x/{entity}` files.

Fixes in commit: https://github.com/noveogroup-amorgunov/nukeapp/commit/15d0fcf57e72ac366d2a86926ad1b678ccc646af

In general, I tried to use eslint rule for public api, but it works a bit weird (mark as error imports inside segments), so I have decided not to use it yet.

illright (approved):
Great stuff!

azinit (approved):
Looks really great! Thank you for contribution ❤️

PR #547, opened 2023-01-22
Add strict definitions for layers

illright:
## Background

We didn't have precise definitions for layers which led to misinterpretations and questions in communities.

## Changelog

1. Added a page about slices in general and each slice specifically
2. Added a page about slices and segments
3. Removed previous pages about units

PR #581, opened 2023-05-17
Add redirects to deleted pages

illright:
## Background

https://github.com/feature-sliced/documentation/pull/547/files

## Changelog

1. Add redirects from deleted pages to new pages


azinit (approved):
Супер! Чекни тольк на netlify-стенде работает ли как нужно

PR #589, opened 2023-06-07
fix: set correct attribute to Discussions section in footer, add translations

AMSHL:
## Background

The link to the "Discussions" section in GitHub in Footer was pointed with the "to" attribute as it's a path inside a router. That can cause some trouble in future versions. And also, because of this, the link was missing the external link icon
image

Also in the footer translations were missing items for the Help and Community links.

## Changelog

 1. Changed link attribute to Discussions section in the Footer
 2. Added translation items for Help and Community links.


azinit (approved):
Thank you for contribution and your attention to detail! ❤️

PR #608, opened 2023-07-30
Docs: update block codes at "Let's form the routes" section

corocoto:
## Background

Update block codes at ["Let's form the routes" section](https://feature-sliced.design/ru/docs/get-started/tutorial#lets-form-the-routes), which allows us to see and use actual version and state of `react-router-dom` library.

### Remark: 

Starting from v6 version,  `react-router-dom`  marked `Switch` and `Redirect` (that placed inside of `Switch`) as deprecated components. As a recomendation `Switch` should be replaced by `Routes`, and `Redirect` by `Navigate`.

### Related links from [`react-router-dom` documentation](https://reactrouter.com/en/main):

* [Upgrade all `<Switch>` elements to `<Routes>`](https://reactrouter.com/en/main/upgrading/v5#upgrade-all-switch-elements-to-routes)
* [Advantages of `<Route element>`](https://reactrouter.com/en/main/upgrading/v5#advantages-of-route-element)
* [Remove `<Redirect>`s inside `<Switch>`](https://reactrouter.com/en/main/upgrading/v5#remove-redirects-inside-switch)

### DIFF
--- a/i18n/en/docusaurus-plugin-content-docs/current/get-started/tutorial.md
+++ b/i18n/en/docusaurus-plugin-content-docs/current/get-started/tutorial.md
@@ -263,7 +263,7 @@ const TestPage = lazy(() => import("./test"));
 export const Routing = () => {
     return (
         <Routes>
-           <Route exact path="/" element={<TestPage/>} />
+           <Route path="/" element={<TestPage/>} />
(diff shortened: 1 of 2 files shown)


azinit:
@feature-sliced/core

illright (approved):

PR #609, opened 2023-07-30
docs: Remove code block, which contained import of common antd styles

corocoto:
## Background

Remove code block, which contained import of `antd/dist/antd.css`. The main reason why I do so - is that this approach doesn't actual at v5 of antd. 

For proving it, you can read [the list of incompatible changes in v5](https://ant.design/docs/react/migration-v5#incompatible-changes-in-v5) on official site, especially ["Technology adjustment" block](https://ant.design/docs/react/migration-v5#technology-adjustment)

> Css files are no longer included in package. Since CSS-in-JS supports importing on demand, the original `antd/dist/antd.css` has also been abandoned. If you need to reset some basic styles, please import `antd/dist/reset.css`.


azinit:
@feature-sliced/core

illright (approved):

PR #634, opened 2023-10-13
feat: add new sample

paghar:
Volt React Dashboard-FSD
It features over 100 UI elements, plugins, and example-based built with React components.


illright (changes requested):
Hi! This project doesn't seem to be following the FSD methodology.

I see four layers, `shared`,`entities`, `pages`, and `app`, which is good, but the layers should consist of [slices and segments](https://feature-sliced.design/docs/reference/slices-segments), which is something that I do not observe in this project.

Also, regarding entities — I'm not sure if this project needs entities at all. Entities represent a real-life concept that an application is dealing with to provide users with business value. This project, if I understood it correctly, is a showcase of components, which by themselves don't have business value until these components are used in an actual application for users.

paghar:
@illright Hi,
Many thanks for your response.

I will change the structure to follow the rules shortly!
Just wanted to know, If I remove the entities layer or rename it to something like `presentation` would that be OK?

illright:
I wouldn't recommend renaming entities to presentation because FSD layer names are standardized. 

This project is a very unusual case, something that I haven't seen other people build with FSD. Usually people use FSD to build websites that are intended for regular people, not other developers.

I would suggest to put everything in Widgets and Pages, maybe with a little bit of Shared, because the business domain in this application is very small, not substantial enough for layers Entities and Features.

paghar:
> I wouldn't recommend renaming entities to presentation because FSD layer names are standardized.
> 
> This project is a very unusual case, something that I haven't seen other people build with FSD. Usually people use FSD to build websites that are intended for regular people, not other developers.
> 
> I would suggest to put everything in Widgets and Pages, maybe with a little bit of Shared, because the business domain in this application is very small, not substantial enough for layers Entities and Features.

Is there any chance to add this project to the FSD examples in the community?

illright:
Of course, but only after all the structural FSD issues are resolved.

Currently, I still don't see the division to slices and segments, only layers with files inside of them. Let me know if any FSD concepts are unclear

paghar:
> Of course, but only after all the structural FSD issues are resolved.
> 
> Currently, I still don't see the division to slices and segments, only layers with files inside of them. Let me know if any FSD concepts are unclear

Slices and Segments are now implemented please take a look [https://github.com/paghar/volt-react-FSD/](https://github.com/paghar/volt-react-FSD/tree/main/src/entities)

as there is no API in this example and it is only presentation oriented, there is only on UI segment available. please let me know if there are more stuff to fix, Thank you.

PR #641, opened 2023-11-21
Rephrase  

fpetrakov:
## Background

Rephrased a sentence in a more natural way


illright (approved):
Hi! That's a good improvement. A new tutorial is on the way, but until then we'll merge this. Thanks!

PR #647, opened 2023-12-17
feat: write the first part of the tutorial in English

illright:
## Background

Continuation of #574 

## Changelog

Replaced the English tutorial with part 1 of the new one. It should be better than the old one even when only consisting of one part.


i18n/en/docusaurus-plugin-content-docs/current/get-started/tutorial.mdx
```diff
@@ -0,0 +1,283 @@
...
+Talking about segments is a good segue into the two layers-exceptions: App and Shared. They are located on top and bottom of the layer stack respectively, and they differ structurally from other layers — they don’t have slices. The Shared layer doesn’t have slices because it’s not tied to any business domain, the App layer doesn’t have slices because it **is** the business domain.
+
+Usually, these two layers are the best places to define your custom segments. Don’t be afraid to create new segments, they don’t cost you anything. For example, your Shared layer could consist of the following segments:
+
+- `ui` — the UI kit
+- `api` — the configuration of the API client
+- `i18n` — locales and internationalization providers
+- `store`\* — store providers
+- `routes` — route constants
+- `env` — environment variables
```
NazariiShvets, line 223: Imo it should be more familiar `config` segment. Since we already declared it in other articles and for envs it could be `shared/config/env.ts`

i18n/en/docusaurus-plugin-content-docs/current/get-started/tutorial.mdx
```diff
@@ -0,0 +1,283 @@
...
+### Business-less layers
+
+Talking about segments is a good segue into the two layers-exceptions: App and Shared. They are located on top and bottom of the layer stack respectively, and they differ structurally from other layers — they don’t have slices. The Shared layer doesn’t have slices because it’s not tied to any business domain, the App layer doesn’t have slices because it **is** the business domain.
+
+Usually, these two layers are the best places to define your custom segments. Don’t be afraid to create new segments, they don’t cost you anything. For example, your Shared layer could consist of the following segments:
+
+- `ui` — the UI kit
+- `api` — the configuration of the API client
+- `i18n` — locales and internationalization providers
+- `store`\* — store providers
```
NazariiShvets, line 221: I don't get this one. Could you elaborate on this, maybe provide some examples?

Which stores we provide? If we talk about some shared providers like I18nProvider or NotificationProvider or something like that, we probably should leave this providers within their original slices

i18n/en/docusaurus-plugin-content-docs/current/get-started/tutorial.mdx
```diff
@@ -0,0 +1,283 @@
...
+- `routes` — route constants
```
NazariiShvets, line 222: > `routes` — route constants

Probably should be `shared/router/paths.ts` rather specific usecase, so if we need add more things there which are about  router it would be clear where to put it

i18n/en/docusaurus-plugin-content-docs/current/get-started/tutorial.mdx
```diff
@@ -0,0 +1,283 @@
...
+
+### Business-less layers
+
+Talking about segments is a good segue into the two layers-exceptions: App and Shared. They are located on top and bottom of the layer stack respectively, and they differ structurally from other layers — they don’t have slices. The Shared layer doesn’t have slices because it’s not tied to any business domain, the App layer doesn’t have slices because it **is** the business domain.
+
+Usually, these two layers are the best places to define your custom segments. Don’t be afraid to create new segments, they don’t cost you anything. For example, your Shared layer could consist of the following segments:
+
+- `ui` — the UI kit
+- `api` — the configuration of the API client
+- `i18n` — locales and internationalization providers
```
NazariiShvets, line 220: > `i18n` — ...

👍

i18n/en/docusaurus-plugin-content-docs/current/get-started/tutorial.mdx
```diff
@@ -0,0 +1,283 @@
...
+  📁 …
+```
+
+### Business-less layers
+
+Talking about segments is a good segue into the two layers-exceptions: App and Shared. They are located on top and bottom of the layer stack respectively, and they differ structurally from other layers — they don’t have slices. The Shared layer doesn’t have slices because it’s not tied to any business domain, the App layer doesn’t have slices because it **is** the business domain.
+
+Usually, these two layers are the best places to define your custom segments. Don’t be afraid to create new segments, they don’t cost you anything. For example, your Shared layer could consist of the following segments:
+
+- `ui` — the UI kit
```
NazariiShvets, line 218: nitpick: Probably should order them in alphabetical order, because they aren't layers, so people don't loose their shit trying to find extensions for ordering or renaming with `01_ui`

```
- api
- config
- i18n
- router
- ui
```
image

i18n/en/docusaurus-plugin-content-docs/current/get-started/tutorial.mdx
```diff
@@ -0,0 +1,283 @@
...
+### Questioning our choices
+
+> *Your scientists were so preoccupied with whether or not they could, they didn’t stop to think if they should.*  
+> — Dr. Ian Malcolm, Jurassic Park, 1993
+
+Let the following realisation sink in:
+
+*Every element that represents as entity or a feature is placed in the folder of that entity or feature.*
+
+Once you fully realise that we’re suggesting to have a separate folder for a lousy Delete button, you may rightfully protest, thinking that it is pure overhead with no benefit. And you’re right, it is a lot of overhead. You don’t have to decompose this precisely, though. We believe in what we like to call **Pain-Driven Development** — when you only need to do something if *not doing it* causes you pain. Many small applications don’t really need to have the Features layer, and sometimes they don’t even need the Entities layer. Our application is a prime example of that, however, for the educational purposes of this tutorial, we will go all the way. <!-- TODO: At the end, we will present two versions of the final application: one *fully decomposed* and one *reasonable*. -->
```
NazariiShvets, line 126: 👍

i18n/en/docusaurus-plugin-content-docs/current/get-started/tutorial.mdx
```diff
@@ -0,0 +1,283 @@
...
+- log in, or present yourself as a certain user
+- log out, or revoke access to a user’s content on a device
+
+The things we write in this list really are just features of our application, only described in excruciating detail. Good thing that Feature-Sliced Design also has a layer called Features! That’s where these will go:
+
+```
+📂 features/
+  📁 edit-note/
+  📁 delete-note/
+  📁 sync-notes/
```
NazariiShvets, line 77: I didn't find any button or something in images which references `sync-notes` feature, so I don't get why we need this feature

i18n/en/docusaurus-plugin-content-docs/current/get-started/tutorial.mdx
```diff
@@ -0,0 +1,283 @@
...
+  📁 authenticate/
+```
+
+We grouped some features together (like create and edit, or log in and log out) and omitted some features altogether (reading notes requires no interaction with the app apart from page navigation) because that’s what makes sense from the technical side.
```
NazariiShvets, line 81: nit: Technicly we didn't group them, we merge them into one feature called `auth`/`authenticate` * (at least I perceived in that way)
Grouping is about new folder without reuse code

Probably should be 
- `authenticate`
   - `sign-out` or `logout`
   - `sign-in`

i18n/en/docusaurus-plugin-content-docs/current/get-started/tutorial.mdx
```diff
@@ -0,0 +1,283 @@
...
+- Shared
+    - UI segment — the UI kit of our application (we should be able to swap this for Ant or MUI if we want to)
+- Entities
+    - Note slice
+        - UI segment — the appearance of a note card
+        - API segment — query for a list of notes from the backend
+        - Model segment — storage of the fetched notes for app-wide access
+- Features
+    - Authenticate slice
+        - UI segment — the button to sign in that displays the current user
```
NazariiShvets, line 257: you missed here mention about sign-in form
image

i18n/en/docusaurus-plugin-content-docs/current/get-started/tutorial.mdx
```diff
@@ -0,0 +1,283 @@
...
+
+Our application’s code will be organized in layers, so let’s recap what each of them contains:
+
+- Shared
+    - UI segment — the UI kit of our application (we should be able to swap this for Ant or MUI if we want to)
+- Entities
+    - Note slice
+        - UI segment — the appearance of a note card
+        - API segment — query for a list of notes from the backend
+        - Model segment — storage of the fetched notes for app-wide access
```
NazariiShvets, line 254: Probably should contain entity user/session so it would be clear from where we read user name `Sign-out (Lev)`

illright:
This tutorial is no good as well

PR #660, opened 2024-03-30
chore: upgrade Docusaurus to 3.3

illright:
## Background

They say it builds faster, also we broke syntax highlighting somehow

## Changelog

1. Upgrade Docusaurus to latest 3.3
2. Fix broken links
3. Fix file names in code blocks that were previously broken


illright:
Review the warnings about potentially broken links

PR #663, opened 2024-05-03
refactor: change order of layer descriptions

Efim-Kapliy:
## Background

Since the image and the description of the layers are arranged in different sequences, this can lead to difficulties for beginners when getting acquainted with the FSD architecture.

## Changelog

1. Changed the order of layer descriptions.
2. Corrected typos in the text.


illright:
Hi, thanks for the PR! I do see the logic in keeping the same order as the image, perhaps it is indeed better that way. Could you also change the English version please?

Efim-Kapliy:
That's it, I've done it. I fixed this for the en-version as well.

illright (approved):

illright:
Thank you!

PR #664, opened 2024-05-04
Update with-nextjs.mdx

yunglocokid:
fix wrong app -> pages


illright (approved):
oops, that's right! thanks :)

PR #671, opened 2024-05-22
docs: added atomic design alternative docs

midas-png:
## Background
I've decided to update Alternatives page on Atomic Design as I know that I could help on this.

## Changelog
1. Updated Atomic Design section in Alternatives page in English and Russian.
2. Deleted WIP notification in Atomic Design section.

Написанную мною секцию проверил на LanguageTool и в Главреде.

По главреду, чистота - **8.7**, читаемость - **8.8**


illright (changes requested):
Hello! Thanks a lot, this is a very important (and long neglected) part of our docs and I'm so glad that you decided to try your hand at it :)

One suggestion I may offer is to reword the headings slightly. I assume they were derived from the notes in the WIP, but those notes are not particularly clear :)

Here are my suggestions:
- Approach -> What is it? / Что это?
- Applicability -> Applicability to frontend / Применимость во фронтенде
- FSD Position -> How does it relate to FSD? / Как оно сочетается с FSD?

I'm also a bit concerned about the part where you write that Atomic Design doesn't have a layer for business logic. Mostly because FSD doesn't have one either :D Instead, FSD offers a separate level of grouping (segments) that creates designated spaces for business logic, data storage, API interactions, app configuration and other things that are relevant for a frontend application, apart from its UI. Would love to see this clarified a bit (and technically, segments can also be applied to Atomic as well, it's just that it doesn't suggest them, since it's mostly a design methodology)

midas-png:
Hi! Thank you for review. Yes, I agree, these names are better than from the WIP. As for a layer for business logic, I agree, that's what I meant, I clarified it

illright (approved):
Alright, good to go! Thanks again :)

PR #672, opened 2024-05-23
Diet app

GorunowMaxim:


illright:
As for the build failure, I think the file extension `.PNG` needs to be lowercased `.png`

illright (changes requested):
Hi, thanks for the patience, I finally got to look at the code. I also tried our upcoming linter on your project to see how it performs, but I noticed that you have some non-standard additions to FSD. Particularly, the concept of subpages and subentities. It's a bit problematic for the following reasons:
1. It's non-standard FSD, so it needs to be prominently documented in the README as project knowledge. Otherwise people will come to this project expecting an FSD structure, but will see something unexpected and get lost.
2. It's an addition to FSD that hurts one of its core benefits — low coupling of sliced. The whole point of slices is to be independent, so in traditional FSD, there's no such thing as a subslice that shares code with its parent slice. It should either just be inside of one slice, or they should be completely separate slices (maybe in a slice group) and cross-imports between them should be avoided.
I would recommend combining all subslices into one slice, and if it gets too big, try to extract something independent from it.

I also noticed that the imports in your project sidestep the public API, for example, `widgets/waterWidget/ui/WaterWidget.tsx` imports `shared/ui/tooltip/Tooltip.tsx`, but reaching into `shared/ui` is prohibited.

Also `features/dailyChallenge` seems empty, it should probably be removed.

And as for the image, you reuploaded it, but it still doesn't show the app's UI, could you please make a screenshot of some page that is not the sign-in?

illright:
Closing this because it got stale

PR #681, opened 2024-06-16
Update decomposition cheatsheet

falkomerr:
Added decomposition by necessity block


illright:
Okay, so, I've been reading this over and over again, I still can't make up my mind about it fully. I don't entirely like it in the current context, because it contradicts the overview. How about I change the overview first, and then we merge this?

illright:
Also I noticed you pushed the SvelteKit guide here, could you extract it in a separate PR please?

falkomerr:
Fine

PR #684, opened 2024-06-25
Feat/sveltekit guide

falkomerr:
Added usage with sveltekit guide


illright (changes requested):
Hi! Thanks for that, seems like it's mostly similar to Nuxt, looks good. A few things:
- the link in the end says Nuxt in both languages
- could you please remove everything else from this PR?

PR #692, opened 2024-07-08
feat: add Quotations App as an example

Egork-Dev:


illright (changes requested):
Hi, thanks for the example! I have a couple of comments about it:
1. It's not clear from the screenshot what the app does, and when I went to the demo, I got fetch errors. I'm not sure the app is working correctly
2. The "Tab" feature does not comply with FSD. It should be a folder with segments and a public API.
3. I suggest running our linter, [Steiger](https://github.com/feature-sliced/steiger), on your code to make sure there aren't any other FSD issues

illright:
This PR has gone stale, so I'll close it, but feel free to re-open it if you'd still like to add this as an example

PR #697, opened 2024-07-24
feat: update decomposition cheatsheet

falkomerr:


illright:
Is there any particular reason that you want this cheatsheet back? I'm hesitant to bring it back even with a disclaimer of some kind because I don't think it's useful in its current form, and it can be rather confusing. Our problem right now is that we have too many definitions that are too narrow, we need to widen the definitions, not constrain them even further

falkomerr:
this cheat sheet really simplifies both the decomposition process for beginners and the overall understanding of what these layers should do

i18n/en/docusaurus-plugin-content-docs/current/get-started/cheatsheet.mdx
```diff
@@ -1,19 +1,31 @@
...
 # Decomposition cheatsheet
 
 Use this as a quick reference when you're deciding how to decompose your UI. PDF versions are also available below, so you can print it out and keep one under your pillow.
 
-## Choosing a layer
+## Decomposition by necessity {#decomposition-by-necessity}
+It is a common mistake to assign separate slides to the part of the code that is not reused anywhere.
+This mistake creates several problems at once:
+- The codebase becomes more disjointed, making the code harder to navigate.
+- A person reading your code, seeing that a slice is not part of the page, may begin to question whether they are breaking part of the display by changing that slice.
```
illright, line 13: issue: it's not clear to me what "display" is

suggestion:
```suggestion
- A person reading your code, seeing that a slice is not part of the page, may begin to question whether this code is used somewhere else, and if they are breaking anything by changing it.
```

i18n/en/docusaurus-plugin-content-docs/current/get-started/cheatsheet.mdx
```diff
@@ -1,19 +1,31 @@
...
+- This is a violation of basic principles such as [KISS](https://en.wikipedia.org/wiki/KISS_principle) and [Occam's Razor](https://en.wikipedia.org/wiki/Occam%27s_razor).
+
+It can be avoided by following a simple rule: **"Parts of code that are not reused anywhere should be where they are used"**.
+
+For example, we have a component that contains a block of information about a user. This component is only used on the profile page, so it should stay on the profile page.
+
+If you are sure that your slice will be used in more than one place, you can use the note below.
```
illright, line 20: suggestion:
```suggestion
If you are sure that your slice will be used in more than one place, or you have another very good reason to decompose the code into the lower layers, you can use the note below.
```

illright (changes requested):
I left a few comments regarding the writing, but I still don't want to bring back that decomposition cheatsheet. It propagates the idea that widgets and pages are just combinations of the layers below, an idea that ended up being quite toxic to the codebase and people's understanding of FSD. There are still some other parts of the docs that propagate the same idea, but I'm working on rewriting them. As for the cheatsheet, I would only be willing to merge this once I rewrite the cheatsheet itself a little bit. I'm not sure when I'll get around to this, but I'll try not to make it too long

illright:
Closing as stale

PR #703, opened 2024-08-03
Feature-Sliced Design 2.1

illright:
## Background

With the new tutorial, the paradigm of FSD has shifted slightly to discourage excessive decomposition and instead treat layers Pages and Widgets as the first stop in searching for code, rather than simply compositional layers. This idea is already widely encouraged by the most active community members, so it's time that we update the docs to reflect that and communicate this paradigm shift by doing a minor version bump.

## Changelog

1. Explicitly permit application-aware things in the Shared layer (things like API and routes, still no business logic)
2. Describe explicitly that imports between segments in Shared and App are allowed

## Release check-list

- [x] The reference pages are all up-to-date
- [x] ~~The About pages are in good shape~~ not necessary
- [x] Steiger supports `@x`
- [x] Steiger 0.5.0 is released
- [x] Suggestion to set up Steiger is added to the migration docs
- [ ] The changelog is written
- [x] The migration guide in the docs is written
- [x] Version bumped


illright:
Released! [Read the full changelog here](https://github.com/feature-sliced/documentation/blob/master/CHANGELOG.md#210---2024-10-31)

PR #705, opened 2024-08-07
Update examples

creator-woow:
Think it's pretty example

 https://github.com/creator-woow/yoldi-profiles


illright:
I noticed that there are errors when I run Steiger, the FSD linter, on this project. Could you fix them please?

creator-woow:
@illright Yeah I see, I'm gonna fix it

creator-woow:
@illright Fixed. Only 4 left, the reason is I can't name pages folder according fsd while using nextjs 14+. As soon as builder sees pages folder it starts build like I use pages router (while I'm using app router)
Снимок экрана 2024-08-12 в 23 48 43

illright:
What about this solution? https://feature-sliced.design/docs/guides/tech/with-nextjs#moving-the-pages-folder-of-nextjs-to-the-root-folder-of-the-project-recommended

creator-woow:
@illright The solution is for pages router, I use latest app router, so it's not fit

creator-woow:
@illright Below was my case solution, I fixed all except these two.  The reason is that they referenced only once, but with app scaling amount on their references will be enough, so I don't thing these are problem
Снимок экрана 2024-08-13 в 19 08 02

illright (changes requested):
These errors point to unnecessary slices. Your app looks quite simple, so it doesn't really need the layer of features (and possibly, doesn't even need entities). Consider trying to fit the app in just Shared, Pages, and App, the architecture will be simpler.

creator-woow:
@illright If I remove unnecessary slices will you merge it?

illright:
These slices are indeed a blocker, but I may have some more comments after this refactor

creator-woow:
@illright Okay, no staiger errors had stayed

creator-woow:
@illright Pls either review & merge, or close pr, if you won't merge it

illright:
I will review it when I get through other higher-priority things and find the time. I'm intending to merge this PR, so I don't want to close it, but I can't give an estimate for when I can do the review. 

If you'd like to speed up the process of getting this PR merged, I can suggest asking someone in our Telegram chat for review. Perhaps they could point out some improvements and you could implement them ahead of the final review

creator-woow:
> I will review it when I get through other higher-priority things and find the time. I'm intending to merge this PR, so I don't want to close it, but I can't give an estimate for when I can do the review.
> 
> If you'd like to speed up the process of getting this PR merged, I can suggest asking someone in our Telegram chat for review. Perhaps they could point out some improvements and you could implement them ahead of the final review

Got it, thank you for fast reply

illright:
I've decided not to stop adding examples to the site because of how much time they take to review and because the example section seems to be a persistent frustration with users. I'd prefer it if in the future we had a completely revised section with examples that would be more curated, but for now, please accept my apology for dragging this PR on for so long and then not merging it.

PR #709, opened 2024-08-13
fix: add/remove website urls in examples

sanua356:
Hi everyone! I looked through all the examples in the documentation and found that some of the website links were broken, and some were in the repository but not listed in the documentation. I fixed that.

- MokeSmoke (add website)
- Sudoky Effector (remove website)

### DIFF
--- a/src/entities/example/index.tsx
+++ b/src/entities/example/index.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback } from "react";
+import React from "react";
 import clsx from "clsx";
 import Image from "@theme/IdealImage";
 import { date } from "@site/src/shared/lib/date";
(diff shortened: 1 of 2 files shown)


illright:
The Sudoku website works for me. Also, could you remove the `/ru` prefix in Moke-Smoke? We don't have internationalization of examples yet, and our docs are English-first

sanua356:
No problem. The "/ru" prefix has been removed from MokeSmoke. But there is a problem with sudoku. It does not open from Russian IPs (only via VPN). Perhaps it is worth leaving a notification about this in the page interface.

illright:
> Perhaps it is worth leaving a notification about this in the page interface.

Ideally, we'd have that, but I wouldn't wanna spend much time on the current examples page, there's a lot to be desired anyway

Could you please fix the linting issues as well?

sanua356:
@illright Yes. I`am fix lint warnings and return sudoku link in config file. As a result, the PR now contains only the addition of the mokeSmoke website to the config file.

illright (approved):

illright:
Thanks for keeping an eye out for this!

PR #712, opened 2024-08-21
Add new v2 project for FSD examples

ifeelBALANCED:


sanua356:
Hello! I tried to register on the site and after sending the forms (registration or login does not matter), I get a message about a redirect. Then nothing happens

ifeelBALANCED:
> Hello! I tried to register on the site and after sending the forms (registration or login does not matter), I get a message about a redirect. Then nothing happens
> 
> 
> 
> 

Hi, the current version of the backend is written in python and deployed to render.com, if no requests have gone there, the backend needs to wake up, which can take time, this is the error of free deployment, redirecting means that the request is still pending, after it passes, you will be redirected. I would look at the site from a computer, just not adapted for mobile yet.

illright (changes requested):
Hi, thanks for the example! I see that there are some Steiger errors, could you investigate please?
```
pnpx steiger ./src
```

illright:
Closing as stale

PR #722, opened 2024-09-20
Update overview.mdx

Moon1Knight:
defintions for 'business domain', 'business entity', 'product feature', and 'business value' are added


illright:
Hi, thanks for the contribution! I like the definitions, but as of now, they don't read very well with the rest of the text. It's kind of an unrelated block in the middle of the text. I'd like to find a way to casually add these definitions to the text or maybe extract them to a reference page and link to it in the overview

illright:
Closing as stale

PR #728, opened 2024-09-29
feat: Japanese full support

ratomiru:
# Description

- This is the PR with all FSD pages high-quality translated to japanese language
- I have passed all pages through chatgpt 4o mini with the reference to main japanese articles about FSD for correct terminology, and then fixed all unnatural texts manually reading it multiple times
- I left only english text links for references because I guess japanese people mainly can not read other languages except japanese and english. If there is a need, it is still possible to look it up switching to other languages for example. Thanks for understanding!

<details><summary>Demo</summary>

スクリーンショット 2024-09-29 18 26 49

</details>

### Question
My japanese company started to use FSD and it is really highly rated! I wonder, maybe it would be good to split "Companies using FSD" section by languages/location to display country local companies? I think other developers/companies will trust FSD more, if it shows them their local country's companies. Of course, there would be a profit for FSD, because there will be a proof that it is used globally. What do you think about it? Thanks!


illright (approved):
Wow, this is huge, thank you so much!

> My japanese company started to use FSD and it is really highly rated! I wonder, maybe it would be good to split "Companies using FSD" section by languages/location to display country local companies? I think other developers/companies will trust FSD more, if it shows them their local country's companies. Of course, there would be a profit for FSD, because there will be a proof that it is used globally. What do you think about it? Thanks!

Yes, this is a good idea. Ideally, we'd want to do it when there would be several companies from each country, but we have to start somewhere, so we might just do it soon. To be honest, it's been a while since we updated that page, since it's not the top priority, but once we get to it, I'll take your suggestion into consideration. Thanks!

ratomiru:
@illright 

> Yes, this is a good idea. Ideally, we'd want to do it when there would be several companies from each country, but we have to start somewhere, so we might just do it soon. To be honest, it's been a while since we updated that page, since it's not the top priority, but once we get to it, I'll take your suggestion into consideration. Thanks!

Got it, thanks!

Also, could you please check algolia settings? It seems that search is currently not working in japanese version

illright:
Sure, I made an issue so that I don't forget about it, I'll let you know once I get it fixed (I hate dealing with Algolia :D)

https://github.com/feature-sliced/documentation/issues/729

PR #734, opened 2024-10-12
upd: Update with-nextjs.mdx

viktorkasap:
Added a note about exporting/importing the page layout with metadata.


illright:
Thanks for bringing it up! Since the StackBlitz playground was created by me, I was able to simply make the changes that you suggested. Is there anything else that you'd like me to change?

viktorkasap:
Thank you for the response.
Great, thanks for responding so quickly.
No, no more comments so far. If I see something that can be improved, I will make a new PR 😊

PR #763, opened 2024-11-21
docs: change usersAdapter to artistAdapter

frontend-sensei:
## Background

Just fixes typo in code example by replacing usersAdapter to artistAdapter
Documentation page:
https://feature-sliced.design/docs/guides/examples/types#how-to-deal-with-nested-dtos

## Changelog

Here, usersAdapter is undefined and probably just wrong. So artistAdapter must be used
image


illright:
Good catch! I noticed also that there's `.users` in that line, which should be `.artists`. Could you change that too, please?

frontend-sensei:
> Good catch! I noticed also that there's `.users` in that line, which should be `.artists`. Could you change that too, please?

Thanks for your response! I noticed too, that in another langs same thing. Fixed.

illright (approved):
Thank you!

PR #768, opened 2024-11-25
Example to v2 list - Project Management System

IvanZemsky:
## Background

Hi there, i've created a project for example list using FSD with react, typescript, mobx, material ui, react-hook-form.
Demo: https://ivanzemsky.github.io/ProjectManagement/
Code: https://github.com/IvanZemsky/ProjectManagement

I launched changed page locally, it seems to be working fine.

## Changelog

- src/pages/examples/_config.ts
- src/pages/examples/img (added project-management-system.png)


illright:
Hi, thanks for the example! I briefly checked the project with Steiger, it's giving an error here — https://github.com/IvanZemsky/ProjectManagement/blob/master/src/entities/Executor/model/store.ts#L8

Looks like `executor` should be `Executor`. I wonder if this is a bug?

IvanZemsky:
> Looks like `executor` should be `Executor`. I wonder if this is a bug?

Huh, yeah, it seems like steiger didn't catch it. I solved the error.

illright:
I've decided not to stop adding examples to the site because of how much time they take to review and because the example section seems to be a persistent frustration with users. I'd prefer it if in the future we had a completely revised section with examples that would be more curated, but for now, please accept my apology for dragging this PR on for so long and then not merging it.

PR #774, opened 2024-12-03
feat: added companies list localization

ratomiru:
## Description

- Added home page companies list localization
  - Locale companies are displayed if there are
  - If there are not enough companies for locale list, some companies are added from default locale
 - Added my japanese company (we use FSD in all new projects)


illright (approved):
That's a wonderful idea! Merging :)

PR #777, opened 2024-12-15
enhancement: Backticks Added for Layers on /get-started

midas-png:
## Background
IMHO, the layer names look nondescript [here](https://feature-sliced.design/docs/get-started/overview#layers).

image

## Changelog
I suggest adding backticks to highlight the layers, just like it’s done on the same page with [segments](https://feature-sliced.design/docs/get-started/overview#segments) and the [basic example](https://feature-sliced.design/docs/get-started/overview#basic-example)

image


illright (changes requested):
Hi! These layer names were deliberately formatted as regular text with capital letters — it's to distinguish the name of the layer and the folder of that layer. This is also why the folders have a folder emoji included.

I do agree that the layer names could stand out a bit more, but I don't think code formatting is the way to go here. Maybe bold would look better?

midas-png:
I agree, bold indeed looks better. I've pushed the commit. Thanks!

image

illright (approved):

illright:
Thanks!

PR #780, opened 2024-12-26
fix: update copyright year 2025 in advance

georgeboozin:
## Background
I noticed several pages/translations have different copyright years.
I've crate PR which could help in 2025.


illright (approved):
Thanks! That makes sense

PR #786, opened 2025-01-07
docs: fsd app replaced with fsd pages in en comments

midas-png:
## Background
Noticed a small issue on the page - https://feature-sliced.design/docs/guides/tech/with-nextjs#app-router, where the `pages` layer is confused with `app` in the comments.
image

## Changelog
Adjusted comment on the page - replaced `FSD app folder` with `FSD pages folder` for `pages` layer


illright (approved):
Good catch, thanks!

PR #787, opened 2025-01-10
docs: layers page translated into Korean

Gaic4o:
## Background

I translated the svelteKit page into Korean.

If you think there are aspects that need improvement, please let me know, and I would be happy to make the adjustments. Thank you!

I have requested a Korean translation review from one reviewer in the FSD Discord Korean community to ensure that my Korean writing is well done. I would appreciate it if you could merge it after the review is completed.


Gaic4o:
I re-examined and revised the whole thing.

i18n/kr/docusaurus-plugin-content-docs/current/reference/layers.mdx
```diff
@@ -0,0 +1,203 @@
+---
+sidebar_position: 1
+pagination_next: reference/slices-segments
+---
+
+# Layer
+
+Layer는 Feature-Sliced Design의 기본 구성 단위입니다. Layer의 주 목적은 코드의 책임 범위와 의존성을 기준으로 코드를 체계적으로 구조화하는 것입니다.
```
illright, line 8: question: how does "Layer는 Feature-Sliced Design의 기본 구성 단위입니다." translate?

Google Translate says it's "Layer is the basic building block of Feature-Sliced ​​Design.", and if so, I think we should clarify that it is the first of the three levels of organizational hierarchy (layer > slice > segment)

Gaic4o: Updated the translation to better reflect the original meaning and hierarchy of Feature-Sliced Design. Clarified that layers are the top level in the Layer > Slice > Segment structure, rather than just 'basic building blocks.' Also refined the wording to highlight how layers define folder structures, helping organize functionality and manage dependencies more effectively.

Gaic4o: @illright How do you feel about the revised parts? I’d appreciate your feedback

i18n/kr/docusaurus-plugin-content-docs/current/reference/layers.mdx
```diff
@@ -0,0 +1,205 @@
...
+pagination_next: reference/slices-segments
+---
+
+# Layer
+
+Layer는 Feature-Sliced Design의 계층 구조에서 가장 상위 수준을 담당하며, Layer > Slice > Segment로 구성된 구조의 첫 번째 단계입니다. 
+Layer의 목적은 프로젝트를 기능별로 정리하고, 모듈 간의 의존성을 효과적으로 관리할 수 있도록 폴더 구조를 분리하는 것입니다. 
+또한, 각 Layer는 특정 기능이 어디에 배치되어야 하는지를 결정하는 기준이 되어, 코드 구조를 체계적으로 유지하는 데 도움을 줍니다.
+
+총 **7개의 Layer**가 있으며, 책임과 의존성이 높은 순서대로 정렬되어 있습니다:
```
illright: issue: this first part doesn't seem right. I translated it with DeepL and Google Translate, seems like the following important aspects are missing:

- Their purpose is to separate code based on how much **responsibility** it needs and **how many other modules** in the app **it depends on**
- There are 7 layers in total, arranged **from most** responsibility and dependency **to least**:

Please feel free to correct me if the automatic translation is at fault here, but I'd like this section to be more closely following the original

illright (changes requested):
praise: great translation! Most sections are translated very well, just the beginning I found a little confusing.

Also, I'm really sorry that it took so long to come around to this PR. I've not been feeling much energy to work on FSD, so PR review got stuffed in the bottom. I really appreciate your contributions though

i18n/kr/docusaurus-plugin-content-docs/current/reference/layers.mdx
Gaic4o: Thank you for the feedback! 
You're correct - I've revised the translation to properly include the key concepts about **responsibility and dependencies** that were missing.

illright (approved):

PR #794, opened 2025-02-03
docs: Standardize technical terms to English in guides/examples Korean documentation

Gaic4o:
## Background

Updated technical terms from Korean to English in the guides/examples pages for consistency and better understanding.

### Changes Made
- Standardized technical terms to English in the following pages:
  - guides/examples/auth.md
  - guides/examples/page-layout.md
  - guides/examples/types.md


i18n/kr/docusaurus-plugin-content-docs/current/guides/examples/auth.md
```diff
@@ -28,9 +28,9 @@ sidebar_position: 1
 
 로그인과 회원가입 컴포넌트를 별도로 만들고, 필요에 따라 index 파일에서 export 할 수 있습니다. 이 컴포넌트들은 사용자로부터 로그인 정보을 입력받는 폼을 포함합니다.
 
-### 로그인 다이얼로그 만들기
+### 로그인 widget 만들기
```
illright, line 31: issue: I think this is not intended — the original title is "Dialog for login", there was no intention to reference the FSD Widget

i18n/kr/docusaurus-plugin-content-docs/current/guides/examples/auth.md
```diff
@@ -90,11 +90,11 @@ export function RegisterPage() {
 
 ## 로그인 정보 전송 방법
 
-로그인 정보를 백엔드 서버로 전송하기 위한 요청 함수를 작성하세요. 이 함수는 상태 관리 라이브러리나 뮤테이션 라이브러리(예: TanStack Query)를 사용하여 호출할 수 있습니다.
+로그인 정보를 백엔드 서버로 전송하기 위한 요청 함수를 작성하세요. 이 함수는 상태 관리 라이브러리나 mutation 라이브러리(예: TanStack Query)를 사용하여 호출할 수 있습니다.
```
illright, line 93: question: why did you include "mutation" in the list of terms? It seems to me like a general concept from computer science that might be more understandable when translated

i18n/kr/docusaurus-plugin-content-docs/current/guides/examples/auth.md
Gaic4o: I think I missed this part, I'll reflect it. Thank you.

i18n/kr/docusaurus-plugin-content-docs/current/guides/examples/auth.md
Gaic4o: You're right. I initially kept "mutation" in English because of TanStack Query's terminology, but you make a good point about it being a general CS concept. 

I suggest we revise it to use the Korean term "데이터 변경" (data modification) which better explains the concept, while still referencing the technical term when specifically talking about the library feature:

`이 함수는 상태 관리 및 데이터 변경(예: TanStack Query의 mutation 기능) 라이브러리를 사용하여 호출할 수 있습니다.`

This makes the concept more accessible while maintaining the connection to the library's technical terminology.

Gaic4o: You raise a good point. "뮤테이션" is just a phonetic translation of "mutation". Instead of this transliteration, we should explain what mutation means in TanStack Query's context.

Gaic4o: Since "mutation" specifically refers to `useMutation`, I think we should explicitly mention this in the example. I suggest adding "(e.g., TanStack Query's useMutation)" to make the technical reference clearer.

What are your thoughts on this suggestion? Did I understand the concept correctly, or am I missing something?

illright: I think "mutation" just means "change". If I would be translating to Russian, I would also translate it to "мутация", which is a phonetic transliteration as well. I don't really see this term as specific to TanStack Query, but I also don't mind if you want to keep it in English instead of transliterated

i18n/kr/docusaurus-plugin-content-docs/current/guides/examples/auth.md
illright: I still see the word "widget", did you forget to push perhaps?

i18n/kr/docusaurus-plugin-content-docs/current/guides/examples/auth.md
Gaic4o: I modified it like this: "This function can be called using state management libraries or mutation features like useMutation from TanStack Query."

Gaic4o: @illright  How do you feel about the revised parts?

illright (approved):

PR #795, opened 2025-02-03
docs: Update Korean title to English in Cross-imports guide

Gaic4o:
## Background

Updated the guide title to use "Cross-import" instead of the Korean term for better clarity and understanding. The English term "Cross-import" is more widely recognized and better represents the technical concept being discussed.


illright (approved):
Fair enough!

PR #823, opened 2025-05-22
docs: add api-requests guide

Solant:
## Background

Some guides on where to place API requests. I added only the English version first to gather feedback.


i18n/en/docusaurus-plugin-content-docs/current/guides/examples/api-requests.md
```diff
@@ -0,0 +1,129 @@
+---
+sidebar_position: 4
+---
+
+# Handling API Requests
+
+Feature-Sliced Design provides a structured way to organize your API request logic.
```
illright: issue: I don't think this statement is very accurate — FSD doesn't provide a way to organize API request logic. I'd say the more accurate description is "here's how to organize request logic with the whole FSD structure"

suggestion: maybe we don't even need this intro and can get straight to the first section?

i18n/en/docusaurus-plugin-content-docs/current/guides/examples/api-requests.md
```diff
@@ -0,0 +1,129 @@
...
+        - 📂 endpoints
+            - 📄 login.ts
+
+The `client.ts` file centralizes your HTTP request setup. It wraps your chosen method (like `fetch()` or an `axios` instance) and handles common configurations, such as:
+
+- Backend base URL.
+- Default headers (e.g., for authentication).
+- Data serialization.
+
+Here are examples for `axios` and `fetch`:
```
illright: suggestion (non-blocking): perhaps it would be clearer if we split the axios and fetch examples into different codeblocks? To demonstrate that they are not meant to be used together

i18n/en/docusaurus-plugin-content-docs/current/guides/examples/api-requests.md
```diff
@@ -0,0 +1,129 @@
...
+
+Organize your individual API request functions in `shared/api/endpoints`, grouping them by the API endpoint.
+
+```ts title="shared/api/endpoints/login.ts"
+import { client } from '../client'; // Assuming 'client' is configured and exported
+
+export interface LoginCredentials {
+  email: string;
+  password: string;
+}
```
illright: suggestion: I believe it's very common nowadays to pair your backend DTO definitions with runtime data validation with Zod or the likes, perhaps we can add something about it here? Maybe a link to [the relevant section of the Types guide](https://feature-sliced.github.io/documentation/docs/guides/examples/types#type-validation-schemas-and-zod) will suffice

i18n/en/docusaurus-plugin-content-docs/current/guides/examples/api-requests.md
illright, line 1: praise: great job with this file! It's concise and to the point :)

i18n/en/docusaurus-plugin-content-docs/current/guides/examples/types.md
```diff
@@ -429,10 +429,6 @@ Other packages simply don't have typings, and you might want to declare them as
 declare module "use-react-screenshot";
 ```
 
-## Auto-generation of types
```
illright: issue: I don't quite like the idea of removing this section entirely. If I was searching for autogeneration of types, I might look in Types or I might look in API requests.

suggestion: I'd prefer if both pages could pick the user up where they are and show them the way. Perhaps we could also do it in a more elaborate way than just having a link — perhaps we can offer some useful clarifications about the content of that link, i. e. how the content of the link applies to the problem at hand.

So basically I would suggest keeping the heading as is, giving a link to the guide, and also saying something like "if you want to auto-generate types from an OpenAPI schema, put them in `shared/api/openapi`. If you want to auto-generate types from a Figma icon library, put them in `shared/ui`.

In fact, maybe we can just paraphrase and not even bother linking, it's just one small paragraph

i18n/en/docusaurus-plugin-content-docs/current/guides/examples/api-requests.md
```diff
@@ -0,0 +1,129 @@
...
+
+:::note
+
+Avoid placing API calls and response types in the `entities` layer prematurely. Backend responses may differ from what your frontend entities need. API logic in `shared/api` or a slice's `api` segment allows you to transform data appropriately, keeping entities focused on frontend concerns.
+
+:::
+
+## Using Client Generators
+
+If your backend has an OpenAPI specification, tools like [orval](https://orval.dev/) or [openapi-typescript](https://openapi-ts.dev/) can generate API types and request functions for you. Place the generated code in `shared/api`.
```
illright: issue: I suspect some people will also want guidance on _where in `shared/api`_ to put the generated code. What was nice about the previous version on another page is that it specified a concrete example — `shared/api/openapi`

suggestion: let's give a specific location here as well, and if they don't like it, they can come up with their own

illright: suggestion: I also like this sentence from the previous version:

> Ideally, you should also include a README in that folder that describes what these files are, how to regenerate them, etc.

Could we add that here as well?

i18n/en/docusaurus-plugin-content-docs/current/guides/examples/auth.md
```diff
@@ -90,60 +90,7 @@ export function RegisterPage() {
...
-```ts title="pages/login/api/login.ts"
-import { POST } from "shared/api";
-
-export function login({ email, password }: { email: string, password: string }) {
-    return POST("/login", { email, password });
-}
-```
-
-You don't have to export the `login()` function in the page's public API, because it's unlikely that any other place in the app will need this request.
+Create a function that makes a request to your backend's login endpoint. This function can either be called directly in the component code using a mutation library (e.g. TanStack Query), or it can be called as a side effect in a state manager: [where to store the request function][examples-api-requests].
```
illright: issue: this link kinda feels patched on top of the text, it doesn't flow naturally, in my opinion.

suggestion: like I wrote in another comment, maybe we could give some clarification about the content of that link, how it's relevant here? Because the linked page is talking about API requests in general, and here we have a very particular API request, so we can offer more precise advice.

I see it as something like "As explained in the [guide for API requests], you can put your request either in `shared/api` or in the `api` segment of your login page."

illright (changes requested):
I really like how the new page reads, great work! I have some concerns about how the other pages link to this one though — I'd like to see the link more naturally integrated into the narrative

i18n/en/docusaurus-plugin-content-docs/current/guides/examples/api-requests.md
Solant: Agree, abstract intro looks fine in theory, but doesn't add any meaningful information

i18n/en/docusaurus-plugin-content-docs/current/guides/examples/api-requests.md
Solant: done

i18n/en/docusaurus-plugin-content-docs/current/guides/examples/api-requests.md
Solant: done

i18n/en/docusaurus-plugin-content-docs/current/guides/examples/types.md
Solant: I rolled this change back to focus on one page at a time, later we can revisit this paragraph if needed

i18n/en/docusaurus-plugin-content-docs/current/guides/examples/api-requests.md
Solant: done

i18n/en/docusaurus-plugin-content-docs/current/guides/examples/auth.md
Solant: done

illright (approved):
Great stuff, thanks!

PR #843, opened 2025-08-05
chore: fix prettier formatting issues

Gaic4o:
## Background
- A Prettier formatting issue was detected in `docusaurus.config.js`, so the file’s code style was updated.


illright (approved):
Ah, I see, that's my fault!

PR #858, opened 2025-08-19
Add Chinese translations for various documentation files including fe…

liuwei1025:
…atures, community, branding, and guides. Introduce new sections for understanding, migration, and tech usage with frameworks like Next.js and Electron.

## Background

Add Chinese translation.


illright (changes requested):
Could you please remove the files that are still in English? Docusaurus provides automatic fallbacks already, so there's no need to keep duplicated content in two locales

liuwei1025:
I have updated all the files to Chinese and verified them locally.

illright:
Thanks! There seems to be a build error still, could you also address that please?

illright (approved):

PR #865, opened 2025-10-03
docs: excessive-entities page

Solant:
## Background

This is a reimagining of the previous MR #859 with previously discussed changes

Some questions before the review:

- I was trying to add a reasoning section about "How excessive entities can harm the project," but it appeared to be harder than I thought (without a good diff view and fs structure widget). Do you think it is needed regardless?
- I'm not 100% satisfied with the code examples I gave, because the majority of them are a bit stretched examples of thin clients. It would be great to have some kind of shared project example with appropriate usage of entities that doesn't require a significant amount of time to be understood. Maybe you have some kind of example?

### DIFF
--- a/i18n/en/docusaurus-plugin-content-docs/current/guides/issues/excessive-entities.md
+++ b/i18n/en/docusaurus-plugin-content-docs/current/guides/issues/excessive-entities.md
@@ -1,0 +1,98 @@
+# Excessive Entities
+
+The entities layer in Feature-Sliced Design is the first layer that incorporates business logic, distinguishing it from the `shared` layer. Unlike the `model` segment, it is globally accessible (except by `shared`), making it reusable across the application. However, its global nature means changes can have a widespread impact, requiring careful design to avoid costly refactors.
+
+## How to keep `entities` layer clean
+
+To keep a maintainable `entities` layer, consider the following principles based on the application's data processing needs. Keep in mind that this classification is not strictly binary, as different parts of the same application may have “thin” or “thick” parts:
+
+- Thin Clients: These applications rely on the backend for most data processing. They often do not require an `entities` layer, as client-side business logic is minimal and involves only data retrieval.
+- Thick Clients: These handle significant client-side business logic, making them suitable candidates for the `entities` layer.
+
+It is acceptable for an application to lack an `entities` layer if it functions as a thin client. This simplifies the architecture and keeps the `entities` layer available for future scaling if needed.
+
+### Avoid Unnecessary Entities
+
+Do not create an entity for every piece of business logic. Instead, leverage types from `shared/api` and place logic in the `model` segment of a current slice. For reusable business logic, use the `model` segment within an entity slice while keeping data definitions in `shared/api`:
+
+```plaintext
+📂 entities
+  📂 order
+    📄 index.ts
+    📂 model
+      📄 apply-discount.ts // Business logic using OrderDto from shared/api
+📂 shared
+  📂 api
+    📄 index.ts
+    📂 endpoints
+      📄 order.ts
+```
+
+### Exclude CRUD Operations from Entities
+
+CRUD operations, while essential, often involve boilerplate code without significant business logic. Including them in the `entities` layer can clutter it and obscure meaningful code. Instead, place CRUD operations in `shared/api`:
+
+```plaintext
+📂 shared
+  📂 api
+    📄 client.ts
+    📄 index.ts
+    📂 endpoints
+      📄 order.ts // Contains all order-related CRUD operations
+      📄 products.ts
+      📄 cart.ts
+```
+
+For complex CRUD operations (e.g., atomic updates, rollbacks, or transactions), evaluate whether the `entities` layer is appropriate, but use it with caution.
+
+### Store Authentication Data in `shared`
+
+Avoid creating a `user` entity for authentication data, such as tokens or user DTOs returned from the backend. These are context-specific and unlikely to be reused outside authentication:
+
+- Authentication responses (e.g., tokens or DTOs) often lack fields needed for broader reuse or vary by context (e.g., private vs. public user profiles).
+- Using entities for auth data can lead to cross-layer imports (e.g., `entities` into `shared`) or usage of `@x` notation, complicating the architecture.
+
+Instead, store authentication-related data in `shared/auth` or `shared/api`:
+
+```plaintext
+📂 shared
+  📂 auth
+    📄 use-auth.ts // Hook returning authenticated user info or token
+    📄 index.ts
+  📂 api
+    📄 client.ts
+    📄 index.ts
+    📂 endpoints
+      📄 order.ts
+```
+
+### Minimize Cross-Imports
+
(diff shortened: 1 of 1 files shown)


i18n/en/docusaurus-plugin-content-docs/current/guides/issues/excessive-entities.md
```diff
@@ -0,0 +1,98 @@
+# Excessive Entities
+
+The entities layer in Feature-Sliced Design is the first layer that incorporates business logic, distinguishing it from the `shared` layer. Unlike the `model` segment, it is globally accessible (except by `shared`), making it reusable across the application. However, its global nature means changes can have a widespread impact, requiring careful design to avoid costly refactors.
```
illright: issue:

> The entities layer in Feature-Sliced Design is the first layer that incorporates business logic

That's not entirely consistent with the rest of the docs, because we mention that it's okay for some business logic to exist in Shared if the size of that business logic doesn't justify a separate layer.

suggestion:

I'm not sure we need the part about how it's distinct from Shared (they are two different layers, after all), so maybe we can shorten to this:

```suggestion
The entities layer in Feature-Sliced Design is one of the lower layers that's primarily for business logic. That makes it widely accessible — all layers except for Shared can access it. However, its global nature means changes to Entities can have a widespread impact, requiring careful design to avoid costly refactors.
```

illright: suggestion: let's also include a little "bridge" phrase between needing to be careful with the design of Entities and the issue of excessive entities, something like:

> One of the most common issues that make the Entities layer problematic is an excess of entities. This can cause ambiguity like "should this file live in the entities, or should it be higher?" and frustration due to import restrictions like "I need this part of the entity in another entity, should I move it to Shared?".

i18n/en/docusaurus-plugin-content-docs/current/guides/issues/excessive-entities.md
```diff
@@ -0,0 +1,98 @@
...
+
+## How to keep `entities` layer clean
+
+To keep a maintainable `entities` layer, consider the following principles based on the application's data processing needs. Keep in mind that this classification is not strictly binary, as different parts of the same application may have “thin” or “thick” parts:
+
+- Thin Clients: These applications rely on the backend for most data processing. They often do not require an `entities` layer, as client-side business logic is minimal and involves only data retrieval.
+- Thick Clients: These handle significant client-side business logic, making them suitable candidates for the `entities` layer.
+
+It is acceptable for an application to lack an `entities` layer if it functions as a thin client. This simplifies the architecture and keeps the `entities` layer available for future scaling if needed.
```
illright: issue: when I was reading this part, I thought that this text **is the advice** on how to keep the Entities layer clean, but turns out that it's just an informational aside, and the actual advice is in the subsequent headings

suggestion: perhaps let's just make this an aside, that way those readers who already know about the distinction of thick vs thin clients could just skip it over:


```suggestion
## How to keep `entities` layer clean

To keep a maintainable `entities` layer, consider the following principles based on the application's data processing needs — it is acceptable for an application to not have an `entities` layer if it functions as a thin client. This simplifies the architecture and keeps the `entities` layer available for future scaling if needed.

:::info[What are thick and thin clients?]

_Thick_ vs. _thin client_ distinction refers to how the application processes data:

- _Thin_ clients rely on the backend for most data processing. Client-side business logic is minimal and involves only exchanging data with the backend.
- _Thick_ clients handle significant client-side business logic, making them suitable candidates for the `entities` layer.

Keep in mind that this classification is not strictly binary, and different parts of the same application may be "thicker" or "thinner" clients.

:::
```

illright: Although I kinda want to have the thick/thin client be the "0th principle" — like "do you even need the entities layer"

i18n/en/docusaurus-plugin-content-docs/current/guides/issues/excessive-entities.md
```diff
@@ -0,0 +1,98 @@
...
+      📄 order.ts // Contains all order-related CRUD operations
+      📄 products.ts
+      📄 cart.ts
+```
+
+For complex CRUD operations (e.g., atomic updates, rollbacks, or transactions), evaluate whether the `entities` layer is appropriate, but use it with caution.
+
+### Store Authentication Data in `shared`
+
+Avoid creating a `user` entity for authentication data, such as tokens or user DTOs returned from the backend. These are context-specific and unlikely to be reused outside authentication:
```
illright: issue: I think this also collides a bit with the guide for authentication, where we present the entities layer as one of the options for storing auth data

suggestion: let's simply recommend storing auth data in shared for simple cases (i.e. most cases)

i18n/en/docusaurus-plugin-content-docs/current/guides/issues/excessive-entities.md
```diff
@@ -0,0 +1,112 @@
...
+- _Thin_ clients rely on the backend for most data processing. Client-side business logic is minimal and involves only exchanging data with the backend.
+- _Thick_ clients handle significant client-side business logic, making them suitable candidates for the `entities` layer.
+
+Keep in mind that this classification is not strictly binary, and different parts of the same application may act as a "thick" or a "thin" client.
+
+:::
+
+### 1. Avoid preemptive slicing
+
+In contrast to previous versions, FSD 2.1 encourages deferred decomposition of slices instead of preemptive, and this approach also extends to `entities` layer. At first, you can place all your code in the `model` segment of your page (widget, feature), and then consider refactoring it later, when business requirements are stable. Remember: it is much easier and safer to move something into `entities` later, than refactor code inside `entities` that can affect any upper level slice functionality.
```
illright: suggestion (non-blocking): just a little rewording for clarity

```suggestion
In contrast to previous versions, FSD 2.1 encourages deferred decomposition of slices instead of preemptive, and this approach also extends to `entities` layer. At first, you can place all your code in the `model` segment of your page (widget, feature), and then consider refactoring it later, when business requirements are stable. 

Remember: the later you move code to the `entities` layer, the less dangerous your potential refactors will be — code in Entities may affect functionality of any slice on higher layers.
```

i18n/en/docusaurus-plugin-content-docs/current/guides/issues/excessive-entities.md
```diff
@@ -0,0 +1,112 @@
...
+    📄 use-auth.ts // authenticated user info or token
+    📄 index.ts
+  📂 api
+    📄 client.ts
+    📄 index.ts
+    📂 endpoints
+      📄 order.ts
+```
+
+### 5. Minimize Cross-Imports
```
illright, line 87: suggestion (non-blocking): would be nice to also link to the Authentication guide here:


```suggestion

For more details on implementing authentication, see [the Authentication guide](/docs/guides/examples/auth).

### 5. Minimize Cross-Imports
```

illright (approved):
It's great, thank you! A few optional suggestions from me, but it's good to go.

<sub>Sorry for taking this long to get to it :p</sub>

Issue #165, opened 2021-07-26
MOTIVATION: Add more comparisons with Clean Architecture


azinit:
https://www.youtube.com/watch?v=qP9biDbPJ0I

martis-git:
https://habr.com/ru/post/499078/

martis-git:
https://habr.com/ru/company/mobileup/blog/335382/

azinit:
Сделать вместе с тикетом:

https://github.com/feature-sliced/documentation/issues/155

azinit:
|              | Clean Architecture                                      | FeatureSliced                                                                                   |   |   |   |   |   |   |   |
|--------------|---------------------------------------------------------|-------------------------------------------------------------------------------------------------|---|---|---|---|---|---|---|
| Конкретность | Очень разный по имплементациям, больше доверия временем | достаточно жестко сформулированный для имплементации, больше про "договоренность разных сторон" |   |   |   |   |   |   |   |
|              |                                                         |                                                                                                 |   |   |   |   |   |   |   |
|              |                                                         |                                                                                                 |   |   |   |   |   |   |   |
|              |                                                         |                                                                                                 |   |   |   |   |   |   |   |
|              | wwdwadaw                                                |                                                                                                 |   |   |   |   |   |   |   |
|              |                                                         |                                                                                                 |   |   |   |   |   |   |   |
|              |                                                         |                                                                                                 |   |   |   |   |   |   |   |
|              |                                                         |                                                                                                 |   |   |   |   |   |   |   |
|              |                                                         |                                                                                                 |   |   |   |   |   |   |   |

Gaic4o:
I’ve been thinking about the idea of adding more comparison examples with Clean Architecture. I think most readers of the official FSD documentation are primarily looking to adopt the methodology or use the documentation as a practical reference for their projects. Of course, comparisons can be helpful when deciding which architectural approach to adopt depending on the project domain, team structure, and other circumstances.

However, I’m not entirely sure how much practical value a direct comparison with Clean Architecture would provide in the FSD documentation. Simply adding comparison examples may not give readers a clear takeaway, and since we haven’t consistently applied Clean Architecture across a wide range of projects ourselves, it may also be difficult to explain the differences, trade-offs, and advantages accurately enough.

So for now, I think it would be better to focus on clearly explaining FSD itself and how to apply it in practice, rather than adding a separate comparison document. For that reason, I’m going to mark this issue as Not planned for now.

PR #904, opened 2026-03-09
docs: add history page 

### LINKED ISSUE #222: Добавить сравнения с существующими подходами (Atomic, DDD, Clean Architecture, ...)
- #204 
- #165 
- #155 
- #59

Solant:
## Background

I updated /docs/about/alternatives/ page to more about history, keeping it in line with original content.

## Changelog

Updated /docs/about/alternatives/ page, URL is the same


Gaic4o (approved):
Looks good to me

Gaic4o:
I re-read the document, and overall it's really well organized. It helped me understand the context in which FSD emerged. I do have one thought on the direction I'd like to share though.

Right now each methodology section reads as though the approach has certain downsides or problems, and therefore FSD is better in that regard. But the way I see it, frontend architecture is an ecosystem that has evolved incrementally by building on and improving previous methodologies, and FSD itself came out of that same flow, learning from and refining what came before.

So rather than highlighting each methodology's shortcomings to emphasize FSD's superiority, I think it might be better to show the difference in context. For example, we could frame it as this methodology being designed for a certain context such as backend or design systems, and FSD bringing that idea into the frontend context in a different way. That way we can respect the other methodologies while still explaining FSD's choices, and more importantly, I think it's simply a more accurate description.

Solant:
From my perspective, there is nothing particularly wrong in being correct and cool and superior, but I see your point. I updated some of the most superior points, and I'll give it a think for a bit in case I come up with something better (that doesn't require adding a ton of text)

Let me know if I got it right now.

Gaic4o:
Since the current title is `History`, I think readers who click on this document may expect to learn when FSD first emerged, how concepts such as FSD’s layers and segments were discussed and developed, and how it evolved into version 2.

However, the current document seems to focus more on how FSD was influenced by existing architectures and methodologies, and how it differs from each of those approaches, rather than on the history of FSD itself. Because of that, I feel there is a slight mismatch between the title and the actual content of the document.

Solant:
Yeah, this is kind of an issue with this page. Previously it was named “Alternatives,” but it was comparing FSD to other “older” approaches, and this name didn't make sense. I tried to rename it to “History” instead to see what happens.

Do you think we can rename it? Something like:
- Alternatives (well, I don't really like it, because those are not really alternatives)
- History (as in “approach evolution history”, not the “FSD” history)
- Comparisons (a new one, but it should be descriptive enough)

Gaic4o:
> Yeah, this is kind of an issue with this page. Previously it was named “Alternatives,” but it was comparing FSD to other “older” approaches, and this name didn't make sense. I tried to rename it to “History” instead to see what happens.
> 
> Do you think we can rename it? Something like:
> 
> * Alternatives (well, I don't really like it, because those are not really alternatives)
> * History (as in “approach evolution history”, not the “FSD” history)
> * Comparisons (a new one, but it should be descriptive enough)

I think going with **Comparisons** would be a good choice.

Solant:
done, let me know if everything is ok now

Gaic4o (approved):

Gaic4o:
Solant if the PR comparing the existing approach is merged in issue #222 would it be okay to close the related issues #59  and #165  without adding any additional comparison cases?

Solant:
> Solant if the PR comparing the existing approach is merged in issue #222 would it be okay to close the related issues #59 and #165 without adding any additional comparison cases?

That's a tough question. Those tickets expect a lot of work to be done across multiple new articles:
- Pick a specific pattern/architecture
- Describe it in details
- Create implementation (as there are no reference implementations for things like ActiveRecord or Clean Architecture for FE)
- Approach the same problem/implementation using FSD
- Document all findings in details

We can link them and close at once if you think this article is good enough, but from my perspective, those tickets require much more deep diving and multiple separate articles.

Gaic4o:
> > Solant if the PR comparing the existing approach is merged in issue #222 would it be okay to close the related issues #59 and #165 without adding any additional comparison cases?
> 
> That's a tough question. Those tickets expect a lot of work to be done across multiple new articles:
> 
> * Pick a specific pattern/architecture
> * Describe it in details
> * Create implementation (as there are no reference implementations for things like ActiveRecord or Clean Architecture for FE)
> * Approach the same problem/implementation using FSD
> * Document all findings in details
> 
> We can link them and close at once if you think this article is good enough, but from my perspective, those tickets require much more deep diving and multiple separate articles.

I agree with you on this.

Looking at the related issues, many of them seem to contain mostly reference links rather than having a clearly discussed direction or scope. I also think they are relatively low priority compared to the documentation work we are currently focusing on.

As you mentioned, properly covering each of these topics would likely require choosing a specific architecture, implementing it, comparing different approaches, and exploring the topic in depth through separate articles.

So, what do you think about closing these issues for now, and creating new issues later with a clearer scope and goal when we actually decide to work on them?

Solant:
Sure, I'm fine with closing them

PR #906, opened 2026-03-14
docs: add slice-groups documentation

Gaic4o:
## Background

Add a dedicated documentation page for **Slice groups**.
This concept was briefly mentioned on the existing **Slices and segments** page, but lacked a more detailed explanation.

### Motivation

The current **Slices and segments** page only introduces slice groups briefly, without explaining when to use them, what rules apply, or what pitfalls to avoid.

Since slice groups are a recurring topic in the FSD community, this dedicated page helps clarify:

- what a slice group is (and what it is not)
- the core rules behind it
- when introducing it makes sense
- concrete examples in entities and pages
- anti-patterns such as placing shared code at the group level

I placed the page under **Reference**, after **Slices and segments** and before **Public API,** because it felt closer to a structural concept than to a usage guide.

This PR includes only the Korean and English versions. I'd prefer to leave translations for the other supported languages to native speakers or more experienced contributors to ensure accuracy.


src/content/docs/docs/reference/slice-groups.mdx
```diff
@@ -0,0 +1,307 @@
...
+A slice group is a folder that groups related slices within the same layer.  
+A slice group itself is not a slice — it does not have its own segments or public API.
+
+In other words, a slice group is a way to make the folder structure easier to find and read without changing FSD's dependency rules.
+
+### Core rules
+
+> **Rule 1: A group folder is just a structural folder**
+>
+> A group folder does not have its own segments like `model`, `ui`, or `api`. It also does not have a public API like `index.ts`.
```
Solant: Let's do a very strict ban on any code reuse instead, so the message is very clear, just like in https://fsd.how/docs/reference/slices-segments/#slice-groups

src/content/docs/docs/reference/slice-groups.mdx
```diff
@@ -0,0 +1,307 @@
...
+</FileTree>
+
+With `order`-related and `customer`-related pages listed flat, you have to scan the entire layer to find pages for a specific domain.
+
+Introducing slice groups allows you to see all pages of the same domain in one place.
+
+<FileTree>
+- pages/
+  - order/ slice group
+    - orderCreate/ slice
```
Solant: nit: do we need to remove prefix part from the grouped slices? like
```
- pages/
  - order/
    - create/
```
or
```
- pages/
  - order/
    - order-create/
```
?

src/content/docs/docs/reference/slice-groups.mdx
```diff
@@ -0,0 +1,307 @@
...
+    - ui/
+  - ...
+</FileTree>
+
+Since `payment/` is not a slice, it does not have its own `model.ts`, `index.ts`, or `api/`.
+
+Not all slices need to be placed in a group.  
+Slices like `user/`, whose meaning is clear on its own, can remain without a group.
+
+### pages — domain-based page grouping
```
Solant: This is definitely new; I've never used this pattern for pages 🤔

src/content/docs/docs/reference/slice-groups.mdx
```diff
@@ -0,0 +1,307 @@
...
+      - ui/
+  - settings/ slice (not a group)
+    - ui/
+  - ...
+</FileTree>
+
+This structure is helpful when there are many pages on the same topic, such as list, detail, create, and edit.
+Opening just one parent folder lets you see all related pages together.
+
+### Can it be used in features too?
```
Solant: I think slice groups were originally introduced to primarily tackle feature layers (I think in 2.0), so it is strange to ban it like this

src/content/docs/docs/reference/slice-groups.mdx
```diff
@@ -0,0 +1,307 @@
...
+
+Yes, it can.  
+However, since features often span multiple entities and multiple user flows, the grouping criteria may be less clear than in entities or pages.
+
+Therefore, it is better to view slice groups in features as an option to be used exceptionally.
+If there is no natural grouping criterion, it is better not to force their introduction.
+
+## Patterns to avoid
```
Solant: I think we need to ban even the idea of doing code share on the slice-group level in the first lines of the document (not literally, but you get the idea).

src/content/docs/docs/reference/slice-groups.mdx
```diff
@@ -0,0 +1,307 @@
...
+
+### Do all slices need to be placed in a group?
+
+No.  
+Only group slices that naturally belong together, and leave the rest as they are.
+
+For example, group only cases where viewing them together is natural, like `payment/invoice` and `payment/receipt`,  
+while slices that are independently understood like `user` and `settings` can remain without a group.
+
+### Is it okay if the group name is included in the import path?
```
Solant: Is it the question regarding absolute vs. relative imports?

Solant (commented):
I have a feeling that this is a bit verbose for the purpose of slice groups

src/content/docs/docs/reference/slice-groups.mdx
Gaic4o: I meant that part as a note about absolute vs. relative imports, but I agree it doesn’t really belong here. I’ll remove it. Thanks for catching that.

src/content/docs/docs/reference/slice-groups.mdx
Gaic4o: Yeah, I agree. I’ll rephrase this more strictly. Also, I think if we include the image from the slice groups section in the docs, we can probably shorten or omit some of the explanation I added here.

src/content/docs/docs/reference/slice-groups.mdx
Gaic4o: I’ll go with `create/` here, since the parent group already provides that context.

src/content/docs/docs/reference/slice-groups.mdx
Gaic4o: I wanted to show that slice groups may also be used in the pages layer. I included this as an example, not as a default pattern.

src/content/docs/docs/reference/slice-groups.mdx
Gaic4o: I see your point. I agree this should not be framed as a ban. As I mentioned before, grouping in the features layer can be less obvious because a feature often spans multiple entities and flows. So rather than treating it as exceptional, I think it would be better to clarify that slice groups can work in some cases but not in others, depending on whether there is a natural grouping criterion. I’m thinking of reflecting that in the document with examples of cases where grouping is a good fit and where it is not.

src/content/docs/docs/reference/slice-groups.mdx
Gaic4o: I see what you mean. I’ll make that explicit in the opening section, and then I think the later explanation can be removed.

Solant (approved):
lgtm

illright (approved):

PR #913, opened 2026-04-01
add Handling Assets page

Solant:
## Background

A small page covering how to handle assets in the application. It is quite a frequent question in FSD discussions, especially about the `assets` segment that doesn't really follow FSD philosophy.

Feedback is welcome, especially about global assets like CSS styles and public assets: I think it is better to move them to the `app` layer, but I'd like to hear your opinions on that.

Please let me know if there are any issues or some topics are not covered.

### DIFF
--- a/src/content/docs/docs/guides/examples/handling-assets.mdx
+++ b/src/content/docs/docs/guides/examples/handling-assets.mdx
@@ -6,17 +6,17 @@ sidebar:
 
 import { FileTree, Aside } from '@astrojs/starlight/components';
 
-Assets are static resources that your application use. In FSD, the placement of assets follows the same principles as code — assets should be grouped based not on the type, but on the use case and should be near the component that uses it.
+Assets are static resources that your application uses. In FSD, assets follow the same placement principles as code. Group them by use case, not by type, and keep them close to the components that use them.
 
 <Aside type="caution">
 
-It's a common practice to implement a [custom `assets` segment](/docs/reference/slices-segments/#segments), but just like the code, this creates an issue of [desegmentation](/docs/guides/issues/desegmented/) and should be generally avoided.
+It's a common practice to implement a [custom `assets` segment](/docs/reference/slices-segments/#segments), but just like with code, this creates an issue of [desegmentation](/docs/guides/issues/desegmented/) and should be generally avoided.
 
 </Aside>
 
 ## Global Assets \{#global-assets}
 
-Global assets include things like global stylesheets (like css reset), fonts and so on. `app` is the appropriate layer for those kind of files, as they are not used by any other component except for the application entrypoint.
+Global assets include things like global stylesheets (such as a CSS reset) and fonts. The `🥞 app` layer is the right place for these files since they're only used by the application entrypoint.
 
 <FileTree>
 - app/
@@ -39,7 +39,7 @@ When an asset is only used by one page, widget, or feature, keep it inside that 
     - index.ts
 </FileTree>
 
-If your UI has a lot of static assets like images, you can create a subfolder in the `ui` segment:
+If your UI has many static assets like images, you can create a subfolder in the `ui` segment:
 
 <FileTree>
 - pages/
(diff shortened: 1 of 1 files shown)


illright (approved):
Looks good, thank you! One non-blocking point is the layer formatting — in the past, I've simply capitalized the name and didn't add any additional formatting or emojis:

> In the App layer, you can find

I don't mind code formatting, but the pancake emoji is a little confusing, and I'm not sure if we explained anywhere what it means

Solant:
Thanks! I'll fix the formatting

I found that some pages use folder emoji to mark some text as folder, so I added pancake emoji for layer names, so they can be distinguished from the other items.

I originally did it in https://fsd.how/docs/guides/tech/with-astro/ , but it I can remove them completely if you think they are misleading. Let me know what you think

illright:
Yeah, I'd be in favor of removing the pancake emoji. The folder emojis can stay — I doubt they will confuse anyone

src/content/docs/docs/guides/examples/handling-assets.mdx
```diff
@@ -0,0 +1,88 @@
...
+  order: 5
+---
+
+import { FileTree, Aside } from '@astrojs/starlight/components';
+
+Assets are static resources that your application uses. In FSD, assets follow the same placement principles as code. Group them by use case, not by type, and keep them close to the components that use them.
+
+<Aside type="caution">
+
+It's a common practice to implement a [custom `assets` segment](/docs/reference/slices-segments/#segments), but just like with code, this creates an issue of [desegmentation](/docs/guides/issues/desegmented/) and should be generally avoided.
```
Gaic4o: I think readers who are unfamiliar with the term may not immediately understand what it means for this to cause a desegmentation issue. It might be helpful to add one or two sentences in the main text explaining why a custom assets segment is discouraged, so readers can understand the context without having to follow the link.

src/content/docs/docs/guides/examples/handling-assets.mdx
```diff
@@ -0,0 +1,88 @@
...
+
+Most bundlers and frameworks provide a `public` folder at the root of your project. Files here are served as-is without processing.
+
+Some build tools like [Vite](https://vite.dev/config/shared-options#publicdir) allow the public folder location to be changed, while others like [Astro](https://docs.astro.build/en/basics/project-structure/#public) don't have this option. If your build tool doesn't support this change, that's fine. The `public` folder is not part of the FSD structure and won't cause naming collisions.
+
+## Summary \{#summary}
+
+| Asset Type | Location |
+|------------|----------|
+| Reusable icons and images | `shared/ui/` |
```
Gaic4o, line 106: The Reusable icons and images → shared/ui/ row in the Summary table feels a bit unclear to me. Since this case is not really explained in the main text, would it make sense to clarify whether this refers to assets used together with shared UI components, or more general-purpose assets reused across different parts of the app?

Gaic4o (approved):

src/content/docs/docs/guides/examples/handling-assets.mdx
Solant: I tried to break the explanation into 2 parts to get a quick answer right away, with optional explanation links. Let me know if it is ok now

src/content/docs/docs/guides/examples/handling-assets.mdx
Solant: I reordered sections to go from local scope to global, and added a section on assets in the `shared` layer. Let me know if it is ok now

src/content/docs/docs/guides/examples/handling-assets.mdx
Gaic4o: Yes, this feels much clearer now.

src/content/docs/docs/guides/examples/handling-assets.mdx
Gaic4o: Yes, the structure feels much more natural now. Thanks!

PR #914, opened 2026-04-02
feat(guide): update tanstack query tech guide

trbetxebz:
Hello, Tanstack Query guide updates :)


illright (changes requested):
Hi, thanks for the PR! I'm a bit concerned about the URL change, and the build is currently failing for that reason as well. I'd prefer it if we kept URLs intact as much as possible — in this case, we can update the title of the page, but keep the old URL. It's an inconsistency, yes, but that looks more preferable to me than having broken links.

P.S. I think the correct casing for "Tanstack" is "TanStack" ([example](https://tanstack.com/router/latest/docs/overview))

trbetxebz:
Return old links. Yeah, TanStack is correct, that was an error in the RU version

illright (approved):

Solant (approved):

PR #922, opened 2026-05-12
remove old docusaurus content, code and dependencies

Solant:


illright (approved):

Gaic4o (approved):

Solant:
@illright what do you think about ESLint? The current ESLint preset uses an unsupported version with an unsupported config format. Besides, eslint-kit uses discontinued plugins (and wasn't updated for 2 years).

I can ignore this for now or migrate it to the modern ESLint with prettier and relevant plugins (but it will have a much smaller ruleset than the current one). What do you think?

illright:
Yeah, feel free to completely overhaul the linting process and modernize it, the current one is trash

illright (approved):

PR #925, opened 2026-05-24
feat: add Lunaria translation status dashboard

Gaic4o:
## Background

- Integrated `@lunariajs/core` to generate a translation status page at `/translation-status/`.
- Added a sidebar link to the page.
- Updated the build process to run `lunaria build` before `astro build`.


Solant:
I think we can use https://lunaria.dev/integrations/starlight/ to directly integrate `lunaria` into `starlight` build without running separate scripts.

astro.config.mjs
```diff
@@ -223,6 +223,11 @@ export default defineConfig({
                         {
                             slug: "docs/branding",
                         },
+                        {
+                            label: "Translation Status",
+                            link: "https://fsd.how/translation-status/",
```
Solant: is it possible to use relative link here?

Gaic4o: Good point, thanks for pointing this out.
A relative link is simpler here and avoids hardcoding the production URL.

Solant (approved):

illright (approved):

PR #930, opened 2026-06-10
fix: always link header logo to the landing page

Gaic4o:
## Background

On translated pages such as `/kr/` and `/ja/` clicking the header logo navigated to the locale-specific `get-started` page instead of the landing page. This happened because Starlight links the logo to the current locale home by default, and the locale home redirects to `get-started`. This overrides the `SiteTitle` component so the header logo always links to the landing page, while keeping the existing redirects unchanged.

https://github.com/user-attachments/assets/b8ffb4f0-f321-40ae-a609-226989bc93f8


Solant:
I have a concern that navigation to the landing page resets the current locale to English

Maybe it is better to implement localized landing pages? In this case we won't have to override the `SiteTitle` component, and locale will be properly preserved.

Gaic4o:
> I have a concern that navigation to the landing page resets the current locale to English
> 
> Maybe it is better to implement localized landing pages? In this case we won't have to override the `SiteTitle` component, and locale will be properly preserved.

This approach makes it much clearer what each item is actually referring to, even with the multi-language setup. It's also a nice way for others to contribute in their own languages.
I'll go ahead and rework it to follow this flow. Thanks!

Solant (approved):

PR #933, opened 2026-07-17
docs: discourage the Widgets layer

Gaic4o:
## Background

This PR updates the FSD documentation to remove the Widgets layer from the set of layers recommended by default.

After discussions with the FSD Core team, we decided that new projects should no longer be encouraged to actively adopt the Widgets layer. This does not mean that the Widgets layer is being removed or prohibited. Existing projects that already use Widgets without significant issues do not need to change their structure. The layer can still be retained when it fits the needs of the project.

The Widgets layer was originally intended for independent UI blocks that compose multiple entities and features. In real frontend applications, however, UI blocks rarely handle presentation alone. They often include data fetching, state management, event handling, and other logic related to user flows. As a result, the responsibilities of the Features layer and the Widgets layer tend to overlap.

This overlap makes it difficult to decide whether a particular piece of code belongs in Features or Widgets. As a project grows, the boundary between the two layers can become even less clear.

For this reason, the updated documentation removes Widgets from the default layer structure. Most code that would previously have been placed in Widgets should instead be organized in Features around user flows and actions. Depending on its responsibility, some code may also belong in App when it is related to application-wide composition or initialization. Code that is reusable and independent of business context may belong in Shared.

There are still cases where Widgets previously provided a natural place for composing multiple Features into a complex UI flow. In these situations, the documentation will recommend considering other composition approaches first. This may include composing the Features in a higher layer, separating responsibilities differently, or exposing interfaces that allow each Feature to be composed externally.

The Widgets related content in the Code Smells & Issues section will remain for now. This content may still be useful for projects that already use the Widgets layer and need guidance on related problems.

Please feel free to leave a comment if you have any questions, suggestions, or concerns about this PR.

### DIFF
--- a/src/content/docs/docs/get-started/faq.mdx
+++ b/src/content/docs/docs/get-started/faq.mdx
@@ -16,10 +16,9 @@ Yes! We have a linter called [Steiger][ext-steiger] to check your project's arch
 
 ### Where to store the layout/template of pages?
 
-Layouts used only for a specific screen can be defined directly in the corresponding `page`. Layouts that are reusable across multiple screens without any business context can be placed in `shared/ui`. Layouts applied to the entire application or to the Route structure should be managed in `app`.
+Layouts used only for a specific screen can be defined directly in the corresponding `page`. Layouts that are reusable across multiple screens without any business context can be placed in `shared/ui`. Layouts applied to the entire application or to the route structure should be managed in `app`.
 
-For more detailed criteria and examples, see the [Page layouts guide][page-layouts].
-
+For more detailed criteria and examples, see the [page layouts guide][page-layouts].
 
 ### What is the difference between a feature and an entity?
 
@@ -29,9 +28,9 @@ For more information, along with examples, see the Reference page on [slices][re
 
 ### Can I embed pages/features/entities into each other?
 
-This is possible, but the composition must happen in a higher Layer. For example, a Page can import multiple `Features` and `Entities` to compose a complete screen. When necessary, components can also be passed through props or children for composition.
+This is possible, but the composition must happen in a higher layer. For example, a page can import multiple features and entities to compose a complete screen. When necessary, components can also be passed through props or children for composition.
 
-However, one `Feature` must not directly import another `Feature` according to the [Layer Import rule][import-rule-layers].
+However, one feature must not directly import another feature according to the [layer import rule][import-rule-layers].
 
 ### What about Atomic Design?
 
--- a/src/content/docs/docs/get-started/faq.mdx
+++ b/src/content/docs/docs/get-started/faq.mdx
@@ -16,10 +16,10 @@ Yes! We have a linter called [Steiger][ext-steiger] to check your project's arch
 
 ### Where to store the layout/template of pages?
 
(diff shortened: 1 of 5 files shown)


Solant:
I think there are some inconsistent naming schemes in those articles, like the capitalization of nouns like Page, quoting them like `Page` (when there is no reference to a layer or specific slice). I think it is better to keep everything the same with the current approach (lower-case nouns like page, layer, widget, etc.) and use quotes only when layer, file name or folder is used (`widgets`, `shared/ui`)

Do you think it's worth revisiting auth and layouts pages later more holistically?

Gaic4o:
> I think there are some inconsistent naming schemes in those articles, like the capitalization of nouns like Page, quoting them like `Page` (when there is no reference to a layer or specific slice). I think it is better to keep everything the same with the current approach (lower-case nouns like page, layer, widget, etc.) and use quotes only when layer, file name or folder is used (`widgets`, `shared/ui`)
> 
> Do you think it's worth revisiting auth and layouts pages later more holistically?

My understanding is that the overview and layout related documentation generally discourages the use of the widgets layer, while `auth.mdx` and `page-layout.mdx` still instruct readers to place code in the widgets layer. This felt inconsistent across the documentation. From a reader’s perspective, it could be confusing to first understand that using widgets is discouraged and then encounter instructions in another document telling them to declare something in the widgets layer.

For that reason rather than revising the overall content of those two documents, I only changed the sections related to widgets in order to reduce this inconsistency. Did you mean that we should keep the existing content of `auth.mdx` and `page-layout.mdx` unchanged in this PR and revisit both documents more holistically later?

Solant:
>Did you mean that we should keep the existing content of auth.mdx and page-layout.mdx unchanged in this PR and revisit both documents more holistically later?

I had this idea, but it is more like a pros and cons solution: it _would_ be great to review it, but I'm not sure how feasible it would be. So this is up to you, I think.

But the capitalization and quoting should be consistent with the rest of the project

Gaic4o:
> > Did you mean that we should keep the existing content of auth.mdx and page-layout.mdx unchanged in this PR and revisit both documents more holistically later?
> 
> I had this idea, but it is more like a pros and cons solution: it _would_ be great to review it, but I'm not sure how feasible it would be. So this is up to you, I think.
> 
> But the capitalization and quoting should be consistent with the rest of the project

Ah, I understand now. I think it would be good to revisit the `auth` and `layouts` documentation more holistically sometime soon.

I should have responded to that point earlier as well, but I missed it. I’ll update the capitalization, backtick usage, and other related details so that they are consistent with the project’s existing style, and then update the PR. Thank you for pointing this out.

Gaic4o:
> > Did you mean that we should keep the existing content of auth.mdx and page-layout.mdx unchanged in this PR and revisit both documents more holistically later?
> 
> I had this idea, but it is more like a pros and cons solution: it _would_ be great to review it, but I'm not sure how feasible it would be. So this is up to you, I think.
> 
> But the capitalization and quoting should be consistent with the rest of the project

I’ve made the changes and applied them.

Solant (approved):

PR #939, opened 2026-07-30
feat: generate social preview images for each page

### LINKED ISSUE #251: PROMO: Add dynamic generation of social preview by article (for specific content)
> По надзадаче #164 

### Description

Сейчас есть ~~(будет)~~ генерация статичной превьюхи для всей доки, а надо бы взять как у apollo с их динамической генерацией

Плюс взвесить все затраты на такое решение

> (стек решений / статей смотри по надзадаче)

### Steps

- [x] PROMO(OG): Add plugin-impl (#368, #387)
- [x] PROMO(OG): Add binding with real template (#375)
- [ ] #400
- [ ] #401
- [ ] #402
- [ ] #404
- [ ] #403

### Reference

image

image

Gaic4o:
## Background

This PR continues the work on generating per-document social preview images that was previously started in #251. Using `astro-og-canvas`, it now automatically generates OG images based on each document’s title and path.

<table>
  <tr>
    <td align="center">
      Preview 1
    </td>
    <td align="center">
      Preview 2
    </td>
    <td align="center">
      Preview 3
    </td>
  </tr>
</table>

## Changelog

Closes #251


package.json
Solant, line 6: Do you think we can skip preview generation for local dev mode using `SKIP_OG=true`?

src/shared/lib/document-id.ts
Solant, line 1: Is it possible to reuse [Astro i18n functions](https://docs.astro.build/en/reference/modules/astro-i18n/) to get canonical document paths and locale segments?

scripts/generate-og-background.mjs
Solant, line 1: Is there a reason to generate a preview manually? I thought that all previews were generated by the `OGImageRoute` endpoint

Solant:
ngl, this looks cool

P.S. Is it possible to make it work for translated pages?

package.json
Gaic4o: As I understand it, `pnpm dev` does not generate all preview images up front. In dev mode, an image is only rendered when its URL is requested, while the build generates all of them.

Do you mean that we should set `SKIP_OG=true` in the `dev` script to disable the OG endpoint entirely during local development, including when an OG image URL is opened?

src/shared/lib/document-id.ts
Gaic4o: That sounds good! Thank you for the suggestion.

scripts/generate-og-background.mjs
Gaic4o: This script doesn’t generate the preview itself. It only generates the shared background image that `OGImageRoute` uses when rendering the actual previews.

Gaic4o:
> ngl, this looks cool
> 
> P.S. Is it possible to make it work for translated pages?

Thanks! I’ve addressed the feedback and added support for translated pages as well.

For Korean, Japanese, and Simplified Chinese, I’m currently using locale-specific Noto Sans fonts as fallbacks, since Roboto Mono doesn’t cover those scripts and Noto Sans keeps the overall look reasonably close. If you think another font would fit the existing OG style better, I’d be happy to switch it.

package.json
Solant: oh, nevermind then

I thought it will generate preview for _every_ page on startup

Solant:
Can you check if the `og:image` value is correct? I get a 404 error when trying to see the image in the preview deployment.

Gaic4o:
> Can you check if the `og:image` value is correct? I get a 404 error when trying to see the image in the preview deployment.

I checked it, and the OG image itself is being generated correctly in the preview deployment. However, `og:image` is currently built as an absolute URL based on `https://fsd.how`, which is configured in `Astro.site`. Because of that, even in the PR preview, it points to the production `/og/...` path. Since this PR has not been deployed to production yet, that URL returns a 404.

The same image can be accessed correctly from the preview deployment origin:

`https://gaic4o-feat-og-images.pr-fsd.pages.dev/og/docs/reference/layers.png`

Once this PR is merged and deployed to production, the `https://fsd.how/og/...` path should work correctly as well.

Solant (approved):

PR #945, opened 2026-08-03
docs: add internationalization (i18n) guide

### LINKED ISSUE #171: Добавить гайд по i18n
От app до shared

image

Gaic4o:
## Background

This guide explains where i18n-related code can be placed in an FSD project.

It focuses on the responsibilities of i18n setup, translation resources, locale, and language switching, including how `shared/i18n`, `app/i18n`, entities, and features can be used, as well as examples of organizing translation resources based on project size and usage scope.

## Changelog

Closes #171

### DIFF
--- a/src/content/docs/docs/guides/examples/Internationalization.mdx
+++ b/src/content/docs/docs/guides/examples/Internationalization.mdx
@@ -1,0 +1,285 @@
+---
+title: Internationalization (i18n)
+sidebar:
+  order: 6
+---
+
+import {
+FileTree,
+Aside,
+Steps,
+Tabs,
+TabItem,
+} from '@astrojs/starlight/components';
+
+Supporting multiple languages involves i18n setup, translation resources, locale management, and language switching.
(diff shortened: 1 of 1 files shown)


Solant:
A bit of feedback from my end:

- nit: i18n setup looks a bit React-specific, especially with i18n instance passing, which might be a bit disruptive for other frameworks like Vue, so I'm not sure if this step is really necessary
- I like the point of not creating i18n as an entity; I would even ban this approach right away
- While I don't really like storing messages and i18n config in `shared/i18n`, it is an understandable approach. but using `i18n` as a slice segment outside `shared` is a bit worrying, as we try to keep custom segments to the minimum (we even had `assets` segment removed in favor of `/ui/assets/`), so I think that we should not make an exception there and use `<slice>/ui/messages` or `<slice>/ui/i18n` for localized messages (just like you did with `/ui/messages.ts`)
- I don't really like the example of `User` entity that uses `Locale`. I think it would be better to make an example of `User` being part of `shared/api` or `shared/auth`, for example

But overall I like this article; great job!

PR #951, opened 2026-08-10
ci: restore the deploy trigger to main

Gaic4o:
## Background

I updated the remaining master references in the deploy workflow and OG image cache restore key to use the current default branch, main. This ensures that both deployment and OG image caching work correctly based on the main branch.


Solant (approved):
I think you can also cleanup algolia tokens from CI, as we no longer use it

Gaic4o:
> I think you can also cleanup algolia tokens from CI, as we no longer use it

I’ve removed it. Thanks for reviewing it.

Solant (approved):

PR #957, opened 2026-09-02
feat: add edit and article request links to the page footer

Gaic4o:
## Background

I created this PR to address #256.

At first, I considered closing the issue as `Not Planned`, but I thought it could still be useful to give readers of the official documentation a more approachable way to suggest topics they would like to see covered.

There are already several channels where people can leave feedback, such as Issues, Discussions, and the community, but I thought having an entry point directly within the documentation could make it a bit easier for readers to naturally turn those thoughts into documentation requests.

So I added `Edit page` and `Request an article` links to the bottom of documentation pages. `Request an article` links to the `Ideas` category in Discussions, where documentation-related ideas are already being shared.

Originally, I wanted to discuss the direction with Solant, who usually reviews these changes, before working on the UI. Since Solant is currently on vacation and the scope of the change is relatively small, I went ahead and put together a quick proposal first. If there is a better way to approach this through Issues or Discussions, I would be happy to discuss it in #256 and adjust the implementation accordingly.

I see this as a small UI addition that provides readers with a slightly easier way to participate in improving the documentation without significantly changing the existing documentation flow.

<p align="center">
  Request an article and Edit page links in the documentation footer
</p>

### DIFF
--- a/src/env.d.ts
+++ b/src/env.d.ts
@@ -1,0 +1,4 @@
+/* `remark-heading-id` ships no type declarations. It reaches TypeScript only
+   because `tsconfig.astro.json` includes `astro.config.mjs`, the file that loads
+   Starlight's `App.Locals` augmentation. */
+declare module "remark-heading-id";
--- a/src/shared/ui/EditLink.astro
+++ b/src/shared/ui/EditLink.astro
@@ -1,0 +1,43 @@
+---
+/* Replaces Starlight's edit link so that both labels stay in English on every
+   locale. Starlight translates its own `page.editLink` string per locale. */
+import { Icon } from "@astrojs/starlight/components";
+
+const requestUrl =
+    "https://github.com/feature-sliced/documentation/discussions/new?category=ideas";
+
+const { editUrl, entry } = Astro.locals.starlightRoute;
+/* Neither link fits the landing page: it is built from the components in
+   `src/pages/_pages/ui/`, so its source file has nothing to edit. */
+const isSplash = entry.data.template === "splash";
+---
+
+{
+    !isSplash && (
+        <>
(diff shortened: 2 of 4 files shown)


src/env.d.ts
Solant, line 1: We can use the FSD-approved way of doing ambient imports: https://fsd.how/docs/guides/examples/types/#ambient-declaration-files-dts

P.S. I'm not 100% fond of the name like `untyped-packages`, as it might create very long paths. (Maybe something like “typings” might be enough, but I'm fine with any folder name.)

src/shared/ui/EditLink.astro
```diff
@@ -0,0 +1,43 @@
...
+{
+    !isSplash && (
+        <>
+            {editUrl && (
+                <a href={editUrl} class="print:hidden">
+                    <Icon name="pencil" size="1.2em" />
+                    Edit page
+                </a>
+            )}
+            <a href={requestUrl} class="print:hidden">
```
Solant, line 24: do you think we need target="_blank" there?

src/shared/ui/EditLink.astro
```diff
@@ -0,0 +1,43 @@
...
+/* Neither link fits the landing page: it is built from the components in
+   `src/pages/_pages/ui/`, so its source file has nothing to edit. */
+const isSplash = entry.data.template === "splash";
+---
+
+{
+    !isSplash && (
+        <>
+            {editUrl && (
+                <a href={editUrl} class="print:hidden">
```
Solant, line 19: 1. Do you think we need target="_blank" there?
2. How is it going to work with someone who has no write access to the repository? From what I can see, the button there is to commit directly into master (which even we can't do)

